PR #4, opened 2016-02-22
Handle primary and foreign keys (task #1100)

georgeconstantinou:


georgeconstantinou:
@mamchenkov can you create a new release?

PR #16, opened 2016-03-24
move Table class logic to a Trait class (task #1391)

georgeconstantinou:


src/CsvMigrationsTableTrait.php
```diff
@@ -0,0 +1,171 @@
...
+        $csvFiles = $this->_getCsvFiles($path);
+        $csvData = $this->_getCsvData($csvFiles);
+
+        if (!empty($csvData)) {
+            foreach ($csvData as $module => $fields) {
+                foreach ($fields as $row) {
+                    if (1 < count($row)) {
+                        $assocModule = $this->_getAssociatedModuleName($row[1]);
+                        if (!empty($assocModule)) {
+                            if ($config['alias'] === $module) {
```
mamchenkov: @georgeconstantinou please refactor ... this is too deep of a nesting ... 

1. When you check for `!empty()` and continue - you can do the opposite: check for `empty()` and exit/return/break.
2. `foreach` loop content can be refactored into a smaller method.

georgeconstantinou:
@mamchenkov please review again

PR #22, opened 2016-03-28
CSV module configuration (task #1610)

georgeconstantinou:


src/ConfigurationTrait.php
```diff
@@ -0,0 +1,90 @@
...
+        }
+    }
+
+    /**
+     * Returns the searchable flag or sets a new one
+     *
+     * @param  bool|null $searchable sets module as (not) searchable
+     * @return bool
+     */
+    public function searchable($searchable = null)
```
mamchenkov: @georgeconstantinou `isSearchable()` is a better name, following the convention of having a verb as the first word in the method name.  It also makes it more readable - it's obvious that the method will return a boolean.  `searchable()` can have a connotation of making something searchable, which this method is not doing.

georgeconstantinou: @mamchenkov fixed, please review again.

flangofas:
:+1:

PR #23, opened 2016-04-04
better constant naming (task #1596)

georgeconstantinou:


src/Table.php
```diff
@@ -16,7 +16,7 @@ class Table extends BaseTable
     /**
      * Searchable parameter name
      */
-    const PARAM_NAME = 'non-searchable';
+    const NON_SEARCHABLE_PARAM = 'non-searchable';
```
mamchenkov: @georgeconstantinou prefixes are better than suffixes, because they make code completion meaningful (usually sorted alphabetically) :)

PR #26, opened 2016-04-07
Added documentation

mamchenkov:


mamchenkov:
@georgeconstantinou please read through this thing as I'm not sure that we have all the features that I'm promising ;)

georgeconstantinou:
@mamchenkov documentation is excellent !! All features mentioned are covered. We also have [these](https://github.com/QoboLtd/cakephp-csv-migrations/blob/documentation/config/bootstrap.php#L4-L7) configuration parameters that can be overridden with the appropriate values, but this I believe should be explained to the plugin's README anyway.

PR #32, opened 2016-04-15
Plugin support for field handlers (task #1787)

georgeconstantinou:


mamchenkov:
@georgeconstantinou maybe it's better to use plugin prefix in the table parameter, if needed, to avoid an extra parameter ... Something like `$table = 'MyFoobarPlugin.Accounts';`.  If that won't work or is too difficult to handle, then we can just merge this implementation.

georgeconstantinou:
@mamchenkov for the related fields in migration files?

mamchenkov:
@georgeconstantinou no, for the field handlers ... like this:

```php
public function renderInput($plugin, $table, $field, $data = '', array $options = [])
```

Instead of adding another parameter `$plugin`, you can still use `$table`.  If you need to specify the table from the plugin, than you can use `PluginName.TableName` notation.

georgeconstantinou:
@mamchenkov got it, will fix.

georgeconstantinou:
@mamchenkov please review

flangofas:
:+1:

PR #46, opened 2016-05-09
Add menus using Events (task #1952)

georgeconstantinou:


leonidsl:
:+1:

src/Template/Element/View/index.ctp
```diff
@@ -63,9 +69,16 @@ if (empty($options['title'])) {
                             </td>
                         <?php endforeach; ?>
                         <td class="actions">
-                            <?= $this->Html->link('', ['action' => 'view', $entity->id], ['title' => __('View'), 'class' => 'btn btn-default glyphicon glyphicon-eye-open']) ?>
-                            <?= $this->Html->link('', ['action' => 'edit', $entity->id], ['title' => __('Edit'), 'class' => 'btn btn-default glyphicon glyphicon-pencil']) ?>
-                            <?= $this->Form->postLink('', ['action' => 'delete', $entity->id], ['confirm' => __('Are you sure you want to delete # {0}?', $entity->id), 'title' => __('Delete'), 'class' => 'btn btn-default glyphicon glyphicon-trash']) ?>
+                            <?php
+                                $event = new Event('View.Index.Menu.Actions', $this, [
```
flangofas: @georgeconstantinou An event wrapper element might be handy here to keep DRY principle.

flangofas:
:+1:

src/Template/Element/View/index.ctp
mamchenkov: @flangofas agreed.  These are some of the quick fixes that we need for the meeting/demo later today, that's why I merged them.  In general, we'll need to find a better way to call events, with less hard-coding of event names and some generic/consistent way for the parameters.

Good catch! ;)

PR #63, opened 2016-05-23
Minor changes (task #1992)

georgeconstantinou:
* Redirect back to view when creating new record
* Type-ahead default options using Configure
* Added type-ahead icon
* Defined format for date, time and date-time field handlers


flangofas:
:+1:

PR #67, opened 2016-05-25
Re-factoring (task #2042)

georgeconstantinou:
### Field Handlers
Moved CSV migration columns handling logic into the Field Handlers.

### Csv Field class
Introduced CsvField class which converts the migration CSV file rows into CSV Field objects.

### Db Field class
Introduced DbField class which converts the CSV Field object to DB Field to be used during the CSV migration.

### Csv File structure
Changed csv files structure (breaks backward compatibility). Dropped field-length column, which now is specified within brackets in the field-type column.

**New structure**
```php
FIELD NAME,FIELD TYPE,REQUIRED,NOT SEARCHABLE
id,uuid
name,string(100),1
description,text
currency,list(currencies)
amount,integer(2)
contact_id,related(CrmRe.Contacts),1
document,file,1
```
**Old structure**
```php
FIELD NAME,FIELD TYPE,FIELD LENGTH,REQUIRED,NOT SEARCHABLE
id,uuid,,,
name,string,100,1,
description,text,,,
currency,list:currencies,,,
amount,integer,2,,
contact_id,related:CrmRe.Contacts,,1,
document,file,,1,
```

### Minor changes
* Re-factored CSV migration Traits.
* CSV migration files are not forced to provide all columns with data, so you can skip adding empty values for optional parameters (such as length, required flag and non-searchable flag):
**New**
```php
description,text
```
**Old**
```php
description,text,,,
```

### DIFF
--- a/src/CsvMigration.php
+++ b/src/CsvMigration.php
@@ -161,7 +161,7 @@ class CsvMigration extends AbstractMigration
     protected function _createFromCsv(array $csvData)
     {
         foreach ($csvData as $col) {
-            $field = $this->_fhf->convertField($col);
+            $field = $this->_fhf->fieldToDb($col);
 
             $this->_table->addColumn($field['name'], $field['type'], [
                 'limit' => $field['limit'],
@@ -194,7 +194,7 @@ class CsvMigration extends AbstractMigration
             if (!in_array($tableFieldName, array_keys($csvData))) {
                 $this->_table->removeColumn($tableFieldName);
             } else {
-                $field = $this->_fhf->convertField($csvData[$tableFieldName]);
+                $field = $this->_fhf->fieldToDb($csvData[$tableFieldName]);
 
                 $this->_table->changeColumn($field['name'], $field['type'], [
                     'limit' => $field['limit'],
--- a/src/CsvMigration.php
+++ b/src/CsvMigration.php
@@ -14,6 +14,7 @@ use Migrations\Table;
 class CsvMigration extends AbstractMigration
 {
     use ConfigurationTrait;
+    use MigrationTrait;
 
     /**
      * File extension
@@ -25,20 +26,6 @@ class CsvMigration extends AbstractMigration
      * @var \Migrations\Table
      */
     protected $_table;
-
-    /**
-     * Field parameters
-     * @var array
-     */
-    protected $_fieldParams = ['name', 'type', 'limit', 'required', 'non-searchable'];
-
-    /**
-     * Error messages
-     * @var array
-     */
-    protected $_errorMessages = [
-        '_createFromCsv' => 'Field parameters count [%s] does not match required parameters count [%s]'
-    ];
 
     /**
      * Field handler factory instance
@@ -71,6 +58,7 @@ class CsvMigration extends AbstractMigration
             $path .= Configure::readOrFail('CsvMigrations.migrations.filename') . '.' . static::EXTENSION;
         }
         $csvData = $this->_getCsvData($path);
+        $csvData = $this->_prepareCsvData($csvData);
         $tableFields = $this->_getTableFields();
 
         if (empty($tableFields)) {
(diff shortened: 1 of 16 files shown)


src/CsvMigration.php
```diff
@@ -199,36 +161,19 @@ protected function _getTableFields()
...
-            }
-            // set to uuid if foreign key
-            if ($this->_isForeignKey($field['type'])) {
-                $field['type'] = 'uuid';
-            }
-            if ($this->_validateField($field)) {
-                $this->_table->addColumn($field['name'], $field['type'], [
-                    'limit' => $field['limit'],
-                    'null' => (bool)$field['required'] ? false : true
+            $csvField = new CsvField($col['name'], $col['type'], $col['required'], $col['non-searchable']);
```
mamchenkov: @georgeconstantinou we can send them all as a single parameter - array, containing all the fields, read from the CSV row and merged with default fields.  This way, in case we change the format later, we won't need to worry about this part and only update the CsvField class handling of the new format.

src/CsvMigration.php
```diff
@@ -199,36 +161,19 @@ protected function _getTableFields()
...
+            $dbField = $this->_fhf->fieldToDb($csvField);
```
mamchenkov: @georgeconstantinou `fieldToDb()` should return an array of fields ... we'll use it later for the combined fields, like `money`, which will return two fields - `amount` and `currency`.

flangofas:
:+1:

georgeconstantinou:
@mamchenkov please review again.

src/CsvTrait.php
```diff
@@ -0,0 +1,53 @@
...
+     *
+     * @param  array  $csvData csv data
+     * @return array
+     */
+    protected function _prepareCsvData(array $csvData)
+    {
+        $result = [];
+        foreach ($csvData as $col) {
+            $col += array_values($this->_defaultParams);
+            $result[$col[0]] = array_combine(array_keys($this->_defaultParams), $col);
```
mamchenkov: @georgeconstantinou can you please rewrite it in a more readable way ... it's not obvious what is coming and coming out of this method :)  it looks like `array_merge()` might be an simpler alternative.

src/CsvTrait.php
```diff
@@ -0,0 +1,53 @@
...
+
+    /**
+     * Method that retrieves csv file data.
+     *
+     * @param  string $path     csv file path
+     * @param  array  $skipRows which rows to skip
+     * @return array            csv data
+     * @todo this method should be moved to a Trait class as is used throught Csv Migrations and Csv Views plugins
+     */
+    protected function _getCsvData($path, array $skipRows = [0])
```
mamchenkov: @georgeconstantinou I don't think you can specify an array with values as a default ... maybe you can, but it pokes my eye and annoys GitHub's syntax highlighting.

I think that if we switch it to `$skipHeaders = true`, it'll be sufficient.

src/FieldHandlers/CsvField.php
```diff
@@ -0,0 +1,260 @@
...
+     * @var bool
+     */
+    protected $_nonSearchable;
+
+    /**
+     * Supported field types
+     *
+     * @var array
+     */
+    protected $_supportedTypes = [
```
mamchenkov: @georgeconstantinou the static list should be replaced with something like `fileHandlerFactory::getList()`.  This way, if we need to support a new field type, we only need to drop the new file handler class into the folder, and then use it in the CSV files, without the need to search and update any other code.

georgeconstantinou: @mamchenkov 73f1a7e

mamchenkov: @georgeconstantinou you implemented the method, but you don't seem to be using it here :)

mamchenkov: @georgeconstantinou sorry ... you are using it, but you still have the `$_supportedTypes` property ... is it needed?

georgeconstantinou: @mamchenkov c03c8cf

src/CsvTrait.php
georgeconstantinou: @mamchenkov 18ad8b9

src/CsvTrait.php
georgeconstantinou: @mamchenkov d675ae0

src/FieldHandlers/CsvField.php
mamchenkov: @georgeconstantinou sorry ... I'm an idiot :)

PR #68, opened 2016-05-26
Throw runtime exception if no csv data are found (task #2042)

georgeconstantinou:
Also list of supported database types is required.

### DIFF
--- a/src/FieldHandlers/DbField.php
+++ b/src/FieldHandlers/DbField.php
@@ -41,6 +41,13 @@ class DbField
     protected $_nonSearchable;
 
     /**
+     * Supported field types
+     *
+     * @var array
+     */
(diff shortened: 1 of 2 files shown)


src/FieldHandlers/DbField.php
```diff
@@ -41,6 +41,13 @@ class DbField
     protected $_nonSearchable;
 
     /**
+     * Supported field types
+     *
+     * @var array
+     */
+    protected $_supportedTypes = ['uuid', 'string', 'integer', 'boolean', 'text', 'datetime', 'date', 'time'];
```
mamchenkov: @georgeconstantinou this is good for now ... later on we'll probably need to find a way to fetch these automatically from CakePHP, or Phinx.  Here's the relevant method from Phinx:

https://github.com/robmorgan/phinx/blob/0.5.x-dev/src/Phinx/Db/Adapter/MysqlAdapter.php#L1144

PR #70, opened 2016-05-27
Created Money field handler (task #2070)

georgeconstantinou:


leonidsl:
:+1:

src/FieldHandlers/MoneyFieldHandler.php
```diff
@@ -0,0 +1,136 @@
...
+        return $result;
+    }
+
+    /**
+     * Method responsible for converting csv field instance to database field instance.
+     *
+     * @param  \CsvMigrations\FieldHandlers\CsvField $csvField CsvField instance
+     * @return array list of DbField instances
+     */
+    public function fieldToDb(CsvField $csvField)
```
flangofas: @georgeconstantinou Sounds like a method of the parent class.

src/FieldHandlers/MoneyFieldHandler.php
```diff
@@ -0,0 +1,136 @@
...
+    /**
+     * Money fields
+     *
+     * @var array
+     */
+    protected $_fields = [
+        'currency' => [
+            'type' => 'string',
+            'field' => 'select',
+            'list' => 'currencies'
```
mamchenkov: @georgeconstantinou this can come from the field definition in CSV ... for example:

```
salary,money(currencies)
tax,money(us_currencies)
```

src/FieldHandlers/MoneyFieldHandler.php
georgeconstantinou: @mamchenkov I am not sure I understand what you mean.

flangofas: @georgeconstantinou converting CSV to DB field method shouldn't be in BaseFieldHandler?

georgeconstantinou: @flangofas there is a standard method to Base field handler, but some handlers handle the conversion differently. For example the [RelatedFieldHandler](https://github.com/QoboLtd/cakephp-csv-migrations/blob/master/src/FieldHandlers/RelatedFieldHandler.php#L145-L161).

src/FieldHandlers/MoneyFieldHandler.php
georgeconstantinou: @mamchenkov see 2842cbb

PR #89, opened 2016-06-14
Check for record by table's unique fields (task #2111)

georgeconstantinou:


georgeconstantinou:
@mamchenkov please re-check

PR #95, opened 2016-06-17
 	Fix multiple level associations (task #2137)

flangofas:
The problem with `$table->alias` is that it takes association name when we deal with associated tables.

As a result, we wouldn't be able to see association when doing the following:

`$this->Projects->Documents->associations();`

So, we replaced it with table which is set in the Table's config as well.

### DIFF
--- a/src/MigrationTrait.php
+++ b/src/MigrationTrait.php
@@ -76,7 +76,7 @@ trait MigrationTrait
      */
     protected function _setAssociationsFromCsv(array $config)
     {
-        $csvData = $this->_csvData();
+        $csvData = $this->_csvData(true);
         $csvObjData = $this->_csvDataToCsvObj($csvData);
         $csvFilteredData = $this->_csvDataFilter($csvObjData, $this->__assocIdentifier);
 
@@ -160,15 +160,23 @@ trait MigrationTrait
     /**
      * Get all modules data.
      *
-     * @return array      Modules, fields and fields types.
+     * @param string $prefix Include plugins name based as per CakePHP's convetion.
+     * @return array         Modules, fields and fields types.
      */
-    protected function _csvData()
+    protected function _csvData($prefix = true)
     {
         $result = [];
         $path = Configure::readOrFail('CsvMigrations.migrations.path');
         $csvFiles = $this->_getCsvFiles($path);
+        $plugin = $this->_getPluginNameFromPath($path);
 
         foreach ($csvFiles as $csvModule => $paths) {
+            if ($prefix) {
+                if (empty($plugin)) {
+                    throw new RuntimeException('Plugin name cannot be prefixed because it was not found.');
+                }
+                $csvModule = $plugin . '.' . $csvModule;
+            }
             foreach ($paths as $path) {
                 $result[$csvModule] = $this->_prepareCsvData($this->_getCsvData($path));
             }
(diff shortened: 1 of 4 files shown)


src/MigrationTrait.php
```diff
@@ -77,60 +76,104 @@ protected function _setAssociationsFromConfig(array $config)
...
+                    unset($fields[$key]);
+                }
+            }
+        }
+
+        return $data;
+    }
+
+    /**
+     * Convert array to object function.
```
mamchenkov: @flangofas this is confusing, since the return value of this method is an array :)

flangofas: @mamchenkov updated.

src/Table.php
```diff
@@ -57,10 +58,9 @@ public function initialize(array $config)
             $this->isSearchable($this->_config['table']['searchable']);
         }
 
-        $this->hasMany('UploadDocuments', [
-            'className' => 'Burzum/FileStorage.FileStorage',
-            'foreignKey' => 'foreign_key',
-        ]);
+        //Set the current module
+        $config['currentMod'] = Inflector::camelize($this->table());
```
georgeconstantinou: This will assume that current module is `Foobar` even if is coming from `MyPlugin.Foobar` and will break if you have `App.Foobar` and `MyPlugin.Foobar`.

src/MigrationTrait.php
```diff
@@ -77,60 +76,104 @@ protected function _setAssociationsFromConfig(array $config)
...
-                     * appending plugin name from current table to associated module.
-                     * @todo investigate more, it might break in some cases, such as Files plugin association.
-                     */
-                    if (!is_null($plugin)) {
-                        $assocModule = $plugin . '.' . $module;
+                } else {
+                    list(, $mod) = pluginSplit($assoccsvModule);
+                    //Foreignkey found in other module
+                    //Let's create hasMany association.
+                    if ($config['currentMod'] === $mod) {
```
georgeconstantinou: If one associated module is `MyPlugin.Foobar` and another associated module is `Foobar` one will override the other. That is why the check was done against `$config['registryAlias']` which includes plugin and module name.

src/Table.php
flangofas: - is this table used in the application level?
- `$config['currentMod']` is [afterwards used against the CSV Module name](https://github.com/QoboLtd/cakephp-csv-migrations/pull/95/files#diff-953835d133d0c96570ab982deb2e55fcR88) which is a single name as well. Shouldn't CSV Module have plugin prefix as well?

mamchenkov: @georgeconstantinou @flangofas we use the `Plugin.Model` notation in the CSVs, so that's what should be returned after parsing.  When comparing, the full model name, including the plugin, should be used as well.  This way we'll avoid any confusion between similar named things in different plugins.

georgeconstantinou: @flangofas 
The problem resides in that `$csvModule` holds the module value as `Foobar` (because is coming from the directory path) and `$assoccsvModule` holds the module value as `MyPlugin.Foobar` (because is coming from the csv files).
That is why previously I was checking in one case against current table's `$config['registryAlias']` and in the other case againt `$config['alias']`

In my opinion to avoid this confusion, the csv modules (`$csvModule`) should include the full name as well. For this to happen, with the current logic, the directory structure should change from: `config/CsvMigrations/migrations/Foobar/migration.csv` to: `config/CsvMigrations/migrations/MyPlugin.Foobar/migration.csv` or find another way to detect if the csv migration files are defined in the Plugin or the App level.

If we get the above logic in place, we should always check against current module (when setting the associations) using it's full name, like `MyPlugin.Foobar`. This will avoid any problems with having modules defined in multiple places, even with the same name.

flangofas:
@mamchenkov @georgeconstantinou please have a look again. Thanks for your feedback!

mamchenkov:
:+1:

georgeconstantinou:
@flangofas excellent stuff :tada:

PR #98, opened 2016-06-22
Pretty values through API (task #2208)

georgeconstantinou:


leonidsl:
:+1:

src/Controller/Api/AppController.php
```diff
@@ -56,6 +67,37 @@ public function view()
             }
         });
 
+        $this->Crud->on('afterFind', function(Event $event) {
+            if (static::FORMAT_PRETTY === $this->request->query('format')) {
```
mamchenkov: @georgeconstantinou rather than having this if in multiple places, just move it into the actual function ... something like this:

```php
// ...
$this->Crud->on('afterFind', function(Event $event) {
  $event->subject()->entity = $this->_prettify($event);
});
```

You'll avoid duplicate code and it'll be more readable as well.

src/Controller/Api/AppController.php
```diff
@@ -56,6 +67,37 @@ public function view()
...
+                $table = $event->subject()->repository->registryAlias();
```
mamchenkov: @georgeconstantinou if you are sending the `$event` of `$entity` as the parameter, the `$table` can be figured out inside the function.  Fewer parameters is better :)

src/Controller/Api/AppController.php
```diff
@@ -56,6 +67,37 @@ public function view()
...
+                $fields = $this->getFieldsDefinitions($event->subject()->repository->alias());
```
mamchenkov: @georgeconstantinou same here...

src/Controller/Api/AppController.php
```diff
@@ -56,6 +67,37 @@ public function view()
...
+                $fhf = new FieldHandlerFactory();
```
mamchenkov: @georgeconstantinou your `view` and `paginate` methods shouldn't know how the prettyfing is done.  Just send the data in and get it back prettyfied.  All the magic should be abstracted into the prettify method.

src/MigrationTrait.php
```diff
@@ -26,11 +26,12 @@
     /**
      * Method that retrieves fields from csv file and returns them in associate array format.
      *
+     * @param  string $moduleName Module Name
      * @return array
      */
-    public function getFieldsDefinitions()
+    public function getFieldsDefinitions($moduleName)
```
mamchenkov: @georgeconstantinou this breaks backward compatibility.  I think it's safe to do so something like:

```php

public function getFieldsDefinitions($moduleName = null) {
  if (empty($moduleName) && is_callable([$this, 'alis'])) {
    $moduleName = $this->alias();
  }
  elseif (empty($moduleName)) {
    throw new \RuntimeException("Failed getting field definitions for unknown module");
  }
}
```

georgeconstantinou: @mamchenkov I updated all calls to this function. The reason I want to avoid `$this->alias()` all together is that it limits the Trait's usage to Table classes.

mamchenkov: @georgeconstantinou I understand.  But given that this is a **public** method, you don't know who and how uses it.  So it's better to provide the safety net.  Alternatively, this is a major release. :)

mamchenkov: @georgeconstantinou you still have a big chunk of code repeated everywhere, where you want to prettyfy the results.  The more API methods will need this, the more copies you will create.  When this will change eventually, there will be a lot of code to update and fix.

If not pushing it into the trait itself, at least make a protected function in the API App Controller, which can be easily called for prettifying results from those controller actions that need it.

georgeconstantinou:
@mamchenkov please re-check

src/Controller/Api/AppController.php
```diff
@@ -136,4 +165,31 @@ public function beforeFilter(Event $event)
...
     }
+
+    /**
+     * Method that prepares entity(ies) to run through pretiffy logic.
+     * It then returns the event object.
+     *
+     * @param  Cake\Event\Event $event Event instance
+     * @return Cake\Event\Event
+     */
+    protected function _prepareToPrettify(Event $event)
```
mamchenkov: @georgeconstantinou the name is confusing ... it suggests that this method only prepares data for prettifying, but in fact it actually does the prettifying.   I know you have a name clash with prettify method, but maybe you should either find a different name (like `prettifyEntity`), or use aliasing, as per http://php.net/manual/en/language.namespaces.importing.php

georgeconstantinou:
@mamchenkov please re-review

PR #108, opened 2016-07-01
Field Handler Factory unit tests

georgeconstantinou:


mamchenkov:
@georgeconstantinou for the qobo-admin-panel, I don't like coupling it like this ... Maybe we can do it differently.  For example:

1. Get the element to use from the rendering options.
2. If the element is specified, use that.
3. If the element is not specified, then check if the qobo-admin-panel element exists.
4. If it exists, use that.
5. If nothing exists or is specified, either throw an exception or render some default way.

What do you think?

georgeconstantinou:
@mamchenkov please re-review.

composer.json
```diff
@@ -8,7 +8,8 @@
         "cakephp/cakephp": ">3.0.0 <4.0",
         "cakephp/migrations": "~1.0",
         "burzum/cakephp-file-storage": "^1.1",
-        "friendsofcake/bootstrap-ui": "~0.3"
+        "friendsofcake/bootstrap-ui": "~0.3",
+        "qobo/cakephp-qobo-admin-panel": "^1.2"
```
mamchenkov: @georgeconstantinou if we don't have this as a requirement now, we can drop the dependency

mamchenkov:
@georgeconstantinou good work mate!

PR #113, opened 2016-07-05
Modify bake template to add skip logic if method is 'create' (task #2244)

georgeconstantinou:


src/Template/Bake/csv_migration.ctp
```diff
@@ -28,9 +28,17 @@ class <%= $name %> extends CsvMigration
      */
     public function change()
     {
+<% if ('create' === $tableMethod): %>
```
mamchenkov: @georgeconstantinou why do you need the `<% ... %>`?

georgeconstantinou: This is part of the template engine, is not visible on the generated file.

georgeconstantinou:
@mamchenkov switched to timestamp, please re-review

src/Shell/Task/CsvMigrationTask.php
```diff
@@ -41,7 +41,7 @@ public function fileName($name)
     {
         $name = $this->getMigrationName($name);
 
-        return $this->__timestamp . '_' . Inflector::camelize($name) . $this->_getUniqueName() . '.php';
+        return $this->__timestamp . '_' . Inflector::camelize($name) . $this->__timestamp . '.php';
```
mamchenkov: @georgeconstantinou you are using the same time stamp twice ... I meant to use the time stamp of the last modification of the `migrations.csv` file.

georgeconstantinou: @mamchenkov please re-review.

PR #116, opened 2016-07-06
Get all records, pagination is handled by jquery DataTables

georgeconstantinou:


mamchenkov:
@georgeconstantinou OK for now, but we should switch the dataTables to use the API ... there might be a lot of records and things will break.

PR #117, opened 2016-07-13
Web UI issues (task #2274)

georgeconstantinou:


mamchenkov:
@georgeconstantinou not the most elegant of the solutions, but OK for now ... We need to find a better way of doing this.

PR #121, opened 2016-07-19
API lookup fields (task #2291)

georgeconstantinou:


src/Table.php
```diff
@@ -135,6 +143,43 @@ public function enablePrimaryKeyAccess()
...
+     * Method that adds lookup fields with the matching values to the Where clause in ORM Query
+     *
+     * @param  \Cake\ORM\Query $query Query instance
+     * @param  string          $id    Record id
+     * @return \Cake\ORM\Query
+     */
+    public function findByLookupFieldsWithValues(Query $query, array $values)
+    {
+        // check for record by table's lookup fields
+        foreach ($this->lookupFields() as $lookupField) {
```
mamchenkov: @georgeconstantinou there is a scenario where `lookupFields()` will return null, if the lookup fields are not set.  Throwing that into the loop will generate a PHP warning, I think.  It's better to get the lookup fields, check that those are not empty and then iterate over them.

PR #124, opened 2016-07-26
View files on edit action - backend (task #2303)

flangofas:


src/FieldHandlers/FileFieldHandler.php
```diff
@@ -84,7 +96,7 @@ protected function _renderInputWithData($table, $field, $data)
      * {@inheritDoc}
      * In this case, it renders img tag or anchor to view the upload files.
      */
-    public function renderValue($table, $field, $data, array $options = [])
+    public function renderValue($table, $field, $data = '', array $options = [])
```
mamchenkov: @flangofas changing the signature of the method, which is defined by the interface is not a good idea ... 

https://github.com/QoboLtd/cakephp-csv-migrations/blob/master/src/FieldHandlers/FieldHandlerInterface.php#L28

PR #129, opened 2016-08-08
Both ends - show/hide panels (task #2322)

flangofas:
**What is the use case:**

When we want to conditionally render panels then we can specify in the config file the following:

In config:

**Structure:**
~~~~
[panels]
Panel Name A = "expression"
~~~~

**Example with values:**
~~~~
[panels]
Company Information = "(%%email%% == 'foobar@foobar.com' && %%location%% == 'other')"
Personal Information = "(%%email%% == 'foobar@foobar.com' || %%country%% == 'Cyprus')"
~~~~

**What is what:**
- Panel is the group of fields defined in the CsvMigrations views
- Expression is conditional expression which holds multiple conditions and their operators

The expression is evaluated using the [Expression Language component](http://symfony.com/doc/current/components/expression_language.html) which returns boolean value.
When the return is true, the panel is shown otherwise the panel is not rendered.


src/Controller/Component/CsvViewComponent.php
```diff
@@ -94,6 +105,70 @@ public function beforeFilter(Event $event)
...
+            }
+        }
+    }
+
+    /**
+     * Parse the config file and set to the _parseConfig.
+     *
+     * @return void
+     */
+    protected function _parseConfig()
```
georgeconstantinou: @flangofas [ConfigurationTrait](https://github.com/QoboLtd/cakephp-csv-migrations/blob/master/src/ConfigurationTrait.php#L67-L74) already handles this and is called in base [Table](https://github.com/QoboLtd/cakephp-csv-migrations/blob/master/src/Table.php#L46) class.

flangofas: @georgeconstantinou ConfigurationTrait is conflicting with InstanceConfigTrait (used in Componet parent class)

src/Controller/Component/CsvViewComponent.php
```diff
@@ -94,6 +106,137 @@ public function beforeFilter(Event $event)
...
+    {
+        //Read module's config
+        $this->_parseConfig();
+        if ($this->_isConditionalPanels()) {
+            $this->panelConditions();
+        }
+    }
+
+    /**
+     * From the given panels in the CsvMigration module config.
```
mamchenkov: @flangofas the recommended doc-block format is:

```
/**
 * Short description of the method (1 line)
 *
 * Long description, separated from short description by an
 * empty line.  Then, after another empty line, all the @-keywords
 * such as @throws, @return, @param, etc.
 *
 * @return void
 */
```

src/Controller/Component/CsvViewComponent.php
georgeconstantinou: @flangofas you don't need to load ConfigurationTrait here. You can access table instance with `$this->_tableInstance` and you can fetch table's csv config with `$this->_tableInstance->getConfig()`

src/Controller/Component/CsvViewComponent.php
```diff
@@ -94,6 +106,137 @@ public function beforeFilter(Event $event)
...
+     * Remove only the specified panels in the config which
+     * their type is not matching with the type field of the entity.
+     *
+     * @return bool|void
+     */
+    public function panelConditions()
```
mamchenkov: @flangofas method names that start with a verb (get, set, is, fetch, apply, etc) are much more readable.  They usually suggest what actions the method will perform and what is the likely return value.

src/Controller/Component/CsvViewComponent.php
```diff
@@ -94,6 +106,137 @@ public function beforeFilter(Event $event)
...
+    {
+        $controller = $this->_registry->getController();
+        $fields = &$controller->viewVars['fields'];
+        $entity = $controller->viewVars['entity'];
+
+        $panels = $this->_parsedConfig[self::PANELS];
```
mamchenkov: @flangofas what happens if there is no panel configuration in the config file? As in, default behavior.

src/Controller/Component/CsvViewComponent.php
flangofas: @georgeconstantinou I see.

src/Controller/Component/CsvViewComponent.php
flangofas: @mamchenkov done!

src/Controller/Component/CsvViewComponent.php
flangofas: @mamchenkov done.

src/Controller/Component/CsvViewComponent.php
flangofas: @mamchenkov See https://github.com/QoboLtd/cakephp-csv-migrations/pull/129/files#diff-b0036d97b3452759d547aa252f152137R115

src/Controller/Component/CsvViewComponent.php
flangofas: @georgeconstantinou done!

src/Controller/Component/CsvViewComponent.php
```diff
@@ -94,6 +105,134 @@ public function beforeFilter(Event $event)
...
+    /**
+     * Compare first against the second value based on the given operator.
+     *
+     * @throws InvalidArgumentException Operator MUST be supported.
+     * @param  string $first    First value
+     * @param  string $operator Comparison operator
+     * @param  string $second   Second value
+     * @return bool
+     */
+    protected function _compare($first, $operator, $second)
```
mamchenkov: @flangofas have a look at this example:

```php
#!/usr/bin/php
<?php

$data = [ 
    'name' => 'Leonid',
    'surname' => 'Mamchenkov',
    'vip' => true,
    'year' => 1978,
];

$expression = '($name == "Leonid" || $year > 1970)';

print_r(evalCondition($expression, $data));

function evalCondition($expression, $data) {
    $result = null;
    
    $expression .= ' ? true : false';
    extract($data);
    eval("\$result = $expression;");
    return $result;
}
```

Obviously, we need to work on it a bit, because `eval()` is a dangerous thing altogether, but if we are doing evaluation of expressions, we need to find the suitable solution.  Either `eval()` or some form of expression parsing, like discussed in [this thread](http://stackoverflow.com/questions/327060/dynamic-logical-expression-parsing-evaluation-in-php).

flangofas: @mamchenkov makes sense. Will look into it.

flangofas: @mamchenkov From what I understand, this implementation need to support all logical operators as well. Correct?

mamchenkov: @flangofas the initial requirement was to support a single field and value condition.  You've started with the logic operators ... :)

mamchenkov:
@flangofas overall, looks good.  But before I merge, I want some unit tests, so that we can test expressions and edge cases to see if the evaluation is done right.  Among other things, I want to see what happens when:

* Field values contain multiple words: `status = 'first attempt'`
* Field values contain special characters: `status = 'attempt #1'`
* Same keyword is used as field and value: `status = 'active' && active = false`

Once we have a unit test with expressions data provider, it'll be easy to throw in some other cases.

webroot/js/panels.js
```diff
@@ -0,0 +1,49 @@
...
+    };
+
+    Panel.prototype.observe = function() {
+        var that = this;
+        var $form = $('form');
+        var url = $form.attr('action');
+        var token = api_options.token;
+        $form.find(':input').change(function() {
+            $.ajax({
+                url: '/api/contacts/panels/',
```
mamchenkov: @flangofas hardcoded URL for just contacts?

flangofas: @mamchenkov Sending more commits.

flangofas: @mamchenkov done.

src/Controller/Api/AppController.php
```diff
@@ -210,6 +211,35 @@ protected function _prependParentModule(ResultSetDecorator $entities)
...
+        $this->request->allowMethod(['ajax', 'post']);
+        $result = [
+            'success' => false,
+            'data' => [],
+        ];
+        $table = $this->loadModel();
+        $tableConfig = $table->getConfig();
+        $entity = $table->newEntity($this->request->data);
+        $panels = Panel::getPanelNames($tableConfig) ?: [];
+        foreach ($panels as $name) {
```
mamchenkov: @flangofas this functionality exists in more than one place ... here and in `beforeRender()` of the CsvViewComponent.php ... should be refactored into a method.

flangofas: @mamchenkov done.

src/Controller/Api/AppController.php
```diff
@@ -210,6 +212,33 @@ protected function _prependParentModule(ResultSetDecorator $entities)
...
+    {
+        $this->request->allowMethod(['ajax', 'post']);
+        $result = [
+            'success' => false,
+            'data' => [],
+        ];
+        $table = $this->loadModel();
+        $tableConfig = $table->getConfig();
+        $entity = $table->newEntity($this->request->data);
+        $panels = Panel::getPanelNames($tableConfig) ?: [];
```
mamchenkov: @flangofas you don't seem to be using this variable anymore ...

flangofas: @mamchenkov Done. Sry, missed it.

tests/TestCase/PanelTest.php
```diff
@@ -0,0 +1,135 @@
...
+        $panel->setName();
+    }
+
+    public function testGetExpression()
+    {
+        $panel = new Panel($this->first, $this->config);
+        $expression = $panel->getExpression();
+
+        //Should not be empty.
+        $this->assertTrue(!empty($expression));
```
mamchenkov: @flangofas `$this->assertFalse(empty($expression));` ... then no need for comment :)

tests/TestCase/PanelTest.php
```diff
@@ -0,0 +1,135 @@
...
+        $config['panels']['Foobar'] = "(%%type%% == 'foobar' && %%name%% == 'antonis')";
+        $panelB = new Panel($panelName, $config);
+
+        $this->assertTrue(in_array('type', $panelB->getFields()), 'Given field name is not in the array');
+        $this->assertTrue(in_array('name', $panelB->getFields()), 'Given field name is not in the array');
+    }
+
+    public function testEvalExpression()
+    {
+        $scenario = 'Scenario 1 - string comparison';
```
mamchenkov: @flangofas please use data provider for scenarios

tests/TestCase/PanelTest.php
flangofas: @mamchenkov done.

tests/TestCase/PanelTest.php
flangofas: @mamchenkov done.

tests/TestCase/PanelTest.php
```diff
@@ -91,45 +88,26 @@ public function testGetFields()
...
-        $data = ['status' => 'attempt #1'];
-        $this->assertTrue($panel->evalExpression($data), sprintf('%s - Expression evaluation failed', $scenario));
-
-        $scenario = 'Scenario 3 - logical operator AND';
         $panelName = 'Foobar';
-        $config['panels']['Foobar'] = "%%status%% == 'active' && %%active%% == false";
+        $config['panels']['Foobar'] = $expression;
         $panel = new Panel($panelName, $config);
-        $data = ['status' => 'active', 'active' => false];
         $this->assertTrue($panel->evalExpression($data), sprintf('%s - Expression evaluation failed', $scenario));
```
mamchenkov: @flangofas it's better to provide the expected value from the data provider.  This way, things can be tested for false values, for example, as well.

flangofas: @mamchenkov done.

PR #130, opened 2016-08-09
Enable API Authentication only if flag is set (task #2211)

georgeconstantinou:


src/Controller/Api/AppController.php
```diff
@@ -55,29 +55,31 @@ public function initialize()
...
-                    'fields' => [
-                        'username' => 'id'
+        if (Configure::read('api_auth')) {
+            // @link http://www.bravo-kernel.com/2015/04/how-to-add-jwt-authentication-to-a-cakephp-3-rest-api/
+            $this->loadComponent('Auth', [
+                // non-persistent storage, for stateless authentication
+                'storage' => 'Memory',
+                'authenticate' => [
+                    // used for validating user credentials before the token is generated
+                    'Form' => [
```
mamchenkov: @georgeconstantinou can you please make sure that the Form authentication still works, even if the API authentication is disabled.

georgeconstantinou: @mamchenkov this only affects the '/api' route, that Form authentication is used when accessing  the `token` method for authenticating the user and generating the token.

PR #131, opened 2016-08-11
Add record id to associated Entity fields, when a lookup field value is used (task #2366)

georgeconstantinou:
Useful for cases where the display field value is used on the associated field. For example a new post is created and in the 'owner' field the username of the user is used instead of its uuid.
```php
BEFORE:
{
    'title' => 'Lorem Ipsum',
    'content' => '.....',
    'owner' => 'admin',
}
```
```php
AFTER:
{
    'title' => 'Lorem Ipsum',
    'content' => '.....',
    'owner' => '77dd9203-3f21-4571-8843-0264ae1cfa48',
}
```

### DIFF
--- a/src/ConfigurationTrait.php
+++ b/src/ConfigurationTrait.php
@@ -71,6 +71,26 @@ trait ConfigurationTrait
         if (file_exists($path)) {
             $this->_config = parse_ini_file($path, true);
         }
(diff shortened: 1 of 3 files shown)


leonidsl:
:+1:

PR #134, opened 2016-08-22
Edit uploaded files (task #2303)

georgeconstantinou:
Relates to https://github.com/QoboLtd/cakephp-qobo-admin-panel/pull/49

### DIFF
--- a/src/Table.php
+++ b/src/Table.php
@@ -47,11 +47,9 @@ class Table extends BaseTable
         //Set the current module
         $config['table'] = $this->_currentTable();
 
-        /**
-         * @todo probably this can be set through migration files, see Files migration.csv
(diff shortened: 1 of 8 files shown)


src/Table.php
```diff
@@ -46,6 +46,16 @@ public function initialize(array $config)
 
         //Set the current module
         $config['table'] = $this->_currentTable();
+
+        /**
+         * @todo probably this can be set through migration files, see Files migration.csv
```
mamchenkov: @georgeconstantinou I think we can do this automagically based on the `files` field type.  So, if the CSV files has something like:
```csv
id,uuid,
title,varchar,
photos,files,
```

then we know that this model will need the Documents relationship and we can dynamically set it and use it.

Not necessarily for this PR though :)

PR #138, opened 2016-08-25
Minor changes on the frontend (task #2322)

flangofas:


webroot/js/panels.js
```diff
@@ -2,10 +2,34 @@
...
 
+    /**
+     * Check if the running action is eligible.
+     *
+     * This function provides the eligible actions for hiding and
+     * showing the panels.
+     * @return {Boolean} True if the current is action is eligible.
+     */
+    Panel.prototype.isEligibleAction = function() {
+        var actions = ['add', 'edit'];
```
mamchenkov: @flangofas this looks a bit too specific to me.  JavaScript evaluation of panels is necessary only when something on the client-side can change the state of the record.  Which is only when you are in the form.  So, if you are in the form, you need to re-evaluate panels on change.  If you are not, you don't.  There might be a variety of actions, which are not add or edit, but which provide a form and require the re-evaluation.

PR #144, opened 2016-08-30
Created database list (task #2417)

flangofas:
### Database list feature:

Let's say from the UI, we have created **contact-types** list and its list items.

Then, in the module's config file we use it as follows:

~~~
FIELD NAME,FIELD TYPE,REQUIRED,NOT SEARCHABLE,UNIQUE
id,uuid
type,dblist(contact-types),
~~~

### What it means:

From the example above, type field is select input field and its options are fetched from the list contact-types. This list must be created by the end-user on the UI first.


config/Migrations/20160830084634_CreateDbLists.php
```diff
@@ -0,0 +1,35 @@
...
+     * @return void
+     */
+    public function change()
+    {
+        $table = $this->table('db_lists', ['id' => false, 'primary_key' => ['id']]);
+        $table->addColumn('id', 'uuid', [
+            'default' => null,
+            'null' => false,
+        ]);
+        $table->addColumn('name', 'string', [
```
mamchenkov: @flangofas list name should probably also be unique

config/Migrations/20160830084645_CreateDbListItems.php
```diff
@@ -0,0 +1,44 @@
...
+            'default' => null,
+            'null' => false,
+            'limit' => 255,
+        ]);
+        $table->addColumn('value', 'string', [
+            'default' => null,
+            'null' => false,
+            'limit' => 255,
+        ]);
+        $table->addColumn('created', 'datetime', [
```
mamchenkov: @flangofas list items should also probably have the order column, to show them in the correct order (sorted by).

src/FieldHandlers/DbListFieldHandler.php
```diff
@@ -0,0 +1,91 @@
...
+{
+    /**
+     * Field type
+     */
+    const FIELD_TYPE = 'string';
+
+    /**
+     * Field type match pattern
+     */
+    const FIELD_TYPE_PATTERN = '/list\((.*?)\)/';
```
mamchenkov: @flangofas this field type pattern is already used for the CSV lists.  You need 'dblist' instead.

config/Migrations/20160830084645_CreateDbListItems.php
flangofas: @mamchenkov at the moment, they are sorted by name in ascending order.

config/Migrations/20160830084634_CreateDbLists.php
flangofas: @mamchenkov done.

mamchenkov:
@flangofas I think we should create the database lists automatically, if they don't already exist.  The admin who will manage the lists does not necessarily know which lists should exist.  So if create them, then it's much easier for him to handle the values.

mamchenkov:
@flangofas I've tested it a bit on my local and it seems to work.  But it'll need more effort.  We should probably merge this PR and do the other work in the additional ones.  Some of the things I've noticed were:

* List value has to be unique.  Otherwise I can add multiple items with the same value and same/different name.
* In the flat list, name has to be unique as well.
* Nested lists are not supported (we do support them in the CSV lists though)
* Active/inactive flag is needed for list items, so that they could be disabled without removing
* The UI needs work.

Waiting for the unit tests from you.

PR #151, opened 2016-09-01
Migrate CSV parsing to league/csv (task #2441)

mamchenkov:
Migrate our own implementation of the CSV parsing to [league/csv](https://packagist.org/packages/league/csv).  As part of this, decouple and refactor the CSV functionality, so that we could get rid of some of the traits.

- [x] Add league/csv to the project
- [x] Add unit tests to compare results from old implementation to the new implementation
- [x] Implement CSV parser classes for migrations, views, lists, etc.
- [x] Replace CsvTrait with CSV parser classes.
- [x] Cleanup
- [x] Test! Test! Test!


src/Parser/Csv/MigrationParser.php
```diff
@@ -0,0 +1,66 @@
...
+     */
+    public function wrapFromPath($path, array $structure = [], $wrapField = null)
+    {
+        $result = [];
+
+        // Overwrite default wrap field
+        if (!empty($wrapField)) {
+            $this->wrapField = $wrapField;
+        }
+
```
flangofas: @mamchenkov You need to overwrite $this->sturcture if necessary here?

src/Parser/Csv/ViewParser.php
```diff
@@ -0,0 +1,92 @@
...
+/**
+ * View CSV Parser
+ *
+ * This parser is useful for view CSV processing.
+ * View CSV files can vary in their format, depending
+ * on the need of the view (for example, we currently
+ * support add/edit/view with panels and index view,
+ * which is completely different.
+ *
+ * Therefor, it is assumed that the caller of the
```
flangofas: @mamchenkov typo

src/Parser/Csv/MigrationParser.php
mamchenkov: @flangofas I'm calling the `parseFromPath()` which does that.

flangofas: @mamchenkov OK :+1:

flangofas:
@mamchenkov Nice! :+1:

PR #155, opened 2016-09-12
Add thumbnails support (task #2486)

georgeconstantinou:


src/FileUploadsUtils.php
```diff
@@ -68,6 +73,13 @@ class FileUploadsUtils
     protected $_fileStorageForeignKey;
 
     /**
+     * Image file extensions
+     *
+     * @var array
+     */
+    protected $_imgExtensions = ['jpg', 'png', 'jpeg'];
```
mamchenkov: @georgeconstantinou GIF is also an image extension

PR #156, opened 2016-09-12
Refactor field handler to rely only to autoloader (task #2493)

mamchenkov:
Field handler factory should rely only on autoloader, and not
traverse through class directories, trying to find files.  If it
does, then field handlers cannot be extended via external libraries.

This commit removes the directory search and replaces a `getList()`
implementation with `hasFieldHandler()` check.

### DIFF
--- a/tests/TestCase/FieldHandlers/FieldHandlerFactoryTest.php
+++ b/tests/TestCase/FieldHandlers/FieldHandlerFactoryTest.php
@@ -94,6 +94,15 @@ class FieldHandlerFactoryTest extends TestCase
         $this->fhf = new FieldHandlerFactory();
     }
 
+    public function testHasFieldHandler()
+    {
+        $result = $this->fhf->hasFieldHandler('string');
+        $this->assertTrue($result, "Failed to find field handler for type 'string'");
+
+        $result = $this->fhf->hasFieldHandler('non-existing-field-type');
+        $this->assertFalse($result, "Found field handler for type 'non-existing-field-type'");
+    }
+
     public function testRenderInput()
     {
         $foos = $this->FooTable->find();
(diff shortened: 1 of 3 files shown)


mamchenkov:
@georgeconstantinou please have a look at this PR.  It looks fine by me and by unit tests, but it's always good to have another pair of eyes on this.  If this is merged, this should be tagged as a major version upgrade as it removes the public static method `getList()` from the `FieldHandlerFactory` class.

tests/TestCase/FieldHandlers/FieldHandlerFactoryTest.php
```diff
@@ -94,6 +94,15 @@ public function setUp()
         $this->fhf = new FieldHandlerFactory();
     }
 
+    public function testHasFieldHandler()
+    {
+        $result = $this->fhf->hasFieldHandler('string');
+        $this->assertTrue($result, "Failed to find field handler for type 'string'");
```
georgeconstantinou: @mamchenkov message should not be _found_ instead of _failed to find_?

tests/TestCase/FieldHandlers/FieldHandlerFactoryTest.php
```diff
@@ -94,6 +94,15 @@ public function setUp()
...
+
+        $result = $this->fhf->hasFieldHandler('non-existing-field-type');
+        $this->assertFalse($result, "Found field handler for type 'non-existing-field-type'");
```
georgeconstantinou: @mamchenkov message should not be _failed to find_ instead of _found_?

tests/TestCase/FieldHandlers/FieldHandlerFactoryTest.php
mamchenkov: @georgeconstantinou the message will be used when the test fails.  So in this case, the test will fail if there is no handler for string field type, and therefor the message will be "Failed to find handler for type 'string'". :)

georgeconstantinou: @mamchenkov cool

PR #157, opened 2016-09-14
Use thumbnails on Views (task #2486)

georgeconstantinou:


src/FieldHandlers/BaseFileFieldHandler.php
```diff
@@ -149,6 +152,15 @@ public function renderValue($table, $field, $data, array $options = [])
             return $result;
         }
 
+        // get file storage image hashes
+        $hashes = Configure::read('FileStorage.imageHashes');
+        if (isset($hashes['file_storage'][$options['imageSize']])) {
+            foreach ($entities as $entity) {
+                $entity->path = dirname($entity->path) . '/' . basename($entity->path, $entity->extension);
```
flangofas: @georgeconstantinou  ```DS``` const?

georgeconstantinou: @flangofas I thought the same, but the path property is the actual url, not a directory path.

PR #160, opened 2016-09-15
API associated records (task #2516)

georgeconstantinou:
API GET requests can now include 1st level associated records by passing an `associated` parameter with value set to `1` or `true`.

Example: `/api/articles/index.json?associated=1`

### DIFF
--- a/src/Controller/Api/AppController.php
+++ b/src/Controller/Api/AppController.php
@@ -449,7 +449,7 @@ class AppController extends Controller
      *
      * @param  Cake\ORM\AssociationCollection $associations       Table associations
      * @param  array                          $nestedAssociations Nested associations
-     * @param  boolean                        $onlyNested         Flag for including only nested associations
+     * @param  bool                           $onlyNested         Flag for including only nested associations
      * @return array
      */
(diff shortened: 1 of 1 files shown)


src/Controller/Api/AppController.php
```diff
@@ -361,4 +387,133 @@ protected function _prettifyEntity(Event $event)
...
+     *
+     * @param  Cake\ORM\AssociationCollection $associations Table associations
+     * @param  array                          $levels       Nested associations
+     * @return array
+     */
+    protected function _containNestedAssociations(AssociationCollection $associations, array $levels)
+    {
+        $result = [];
+        foreach ($associations as $association) {
+            if (current($levels) !== $association->className()) {
```
flangofas: @georgeconstantinou don't like the idea of using internal pointers of a different array in a ```foreach``` loop because ```foreach``` loops tend to manipulate internal pointers in the background. Maybe ```$levels``` should be an instance of ```ArrayIterator``` ?

georgeconstantinou: :+1:

PR #165, opened 2016-09-16
	UI improvements (task #2417)

flangofas:
### UI Improvements
- [x] Create list items using list ID
- [x] UI improvements
- [x] Change position of list items from UI
- [x] Active/inactive flag on list items

### Further Improvements as per comment
- [x] Move 'Add item' button from the dblists index to the dblist-items index (top right, like the other pages)
- [x] Show both name and value in the dblist-items index
- [x] Show active/inactive in the dblist-items inde, so that those items can be re-actived/deleted, etc.
- [x] Add 'Back' button in the dblist-items (top right, next to 'Add item' button), so that the user can easily navigate back to the index of dblists.

### Dblist index
ss-dblist-index

### DblistItem index
ss-dblist-items-index

### DblistItem edit
ss-dblist-items-edit


mamchenkov:
@flangofas so far so good :)

mamchenkov:
@flangofas much better now.  I think the following changes should improve the UI even further:

* Move 'Add item' button from the dblists index to the dblist-items index (top right, like the other pages)
* Show both name and value in the dblist-items index
* Show active/inactive in the dblist-items inde, so that those items can be re-actived/deleted, etc.
* Add 'Back' button in the dblist-items (top right, next to 'Add item' button), so that the user can easily navigate back to the index of dblists.

flangofas:
@mamchenkov updated with screenshots

mamchenkov:
@flangofas 

1. for the 'Add item' and 'Back' buttons, you can use the icons. I think you already did use the icon on the 'Add item' before, when it was in the dblists index.
2. for the dblist item edit, no need for the 'Add item' button, but a 'Back' button might be useful.

flangofas:
@mamchenkov 
>  dblist item edit, no need for the 'Add item' button, but a 'Back' button might be useful.

Done.

>  for the 'Add item' and 'Back' buttons, you can use the icons. I think you already did use the icon on the 'Add item' before, when it was in the dblists index.

Will do it in the next PR. I am merging this one.

PR #166, opened 2016-09-16
Simplify panel tests

mamchenkov:


flangofas:
@mamchenkov  makes more sense :smile:

PR #185, opened 2016-09-29
Minor UI improvements

mamchenkov:
* Better display of the related record links (using labels)
* Fix mistakenly escaped HTML in the fields of the associated records
* Add plus icon on the 'Add' button of the index top menu

### DIFF
--- a/src/Events/ViewMenuListener.php
+++ b/src/Events/ViewMenuListener.php
@@ -53,9 +53,10 @@ class ViewMenuListener implements EventListenerInterface
         $urlAdd = ['plugin' => $request->plugin, 'controller' => $request->controller, 'action' => 'add'];
 
         $btnAdd = $appView->Html->link(
+            '<span class="glyphicon glyphicon-plus"></span> ' .
             __('Add {0}', Inflector::singularize($options['title'])),
             $urlAdd,
-            ['class' => 'btn btn-primary']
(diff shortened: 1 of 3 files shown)


src/Events/ViewMenuListener.php
```diff
@@ -53,9 +53,10 @@ public function getIndexMenuTop(Event $event, Request $request, array $options)
         $urlAdd = ['plugin' => $request->plugin, 'controller' => $request->controller, 'action' => 'add'];
 
         $btnAdd = $appView->Html->link(
+            '<span class="glyphicon glyphicon-plus"></span> ' .
```
flangofas: @mamchenkov 
If you prefer:

```php
$view->Html->icon('plus')
```

mamchenkov: @flangofas I don't see the `icon()` method on the `HtmlHelper` ... where is it coming from?

mamchenkov: @flangofas sounds useful, but doesn't work.  The `$appView` variable doesn't seem to have the BootstrapUI helper loaded.  We'll need to look into it in a separate PR.

Thanks for the pointer!

flangofas: @mamchenkov done. #189

mamchenkov: @flangofas Thanks! Awesome! :+1:

PR #186, opened 2016-09-30
Switch views pagination to datatables (task #2629)

georgeconstantinou:
:warning: **Backward compatibility breaker**

On index View no records are fetched server-side, but instead, they are fetched using API call from DataTables library. This brought changes all over the plugin's code.

On API interface the Auth component is now always enabled but will allow access to all actions if the API authentication flag is _off_. This allows the following logic to work:
If API authentication is set to _on_, access is restricted to `token()` action for validating the user and generating an API token. If API authentication is set to _off_, access is allowed to all API actions (not recommended, but optional for systems where IP access restrictions are in place and want to skip API authentication). The plugin always generates an API token for the system's currently logged in user. This token is passed to all ajax calls made within the system (for example the DataTables library mentioned above), so the user can be verified and used by any third party ACL libraries.

In addition, API responses can now include UI menus, such as action buttons on the index view (view, edit, delete), which again can be filtered through third party ACL libraries.

### DIFF
--- a/src/Controller/AppController.php
+++ b/src/Controller/AppController.php
@@ -52,7 +52,6 @@ class AppController extends BaseController
     public function index()
     {
         $this->render('CsvMigrations.Common/index');
-        $this->set('_serialize', ['entities']);
     }
 
     /**
--- a/src/Controller/AppController.php
+++ b/src/Controller/AppController.php
@@ -51,7 +51,6 @@ class AppController extends BaseController
      */
     public function index()
     {
-        $this->set('entities', $this->paginate($this->{$this->name}));
         $this->render('CsvMigrations.Common/index');
         $this->set('_serialize', ['entities']);
     }
--- a/src/Controller/AppController.php
+++ b/src/Controller/AppController.php
@@ -14,7 +14,10 @@ class AppController extends BaseController
     public function initialize()
     {
         parent::initialize();
+
         $this->_fileUploadsUtils = new FileUploadsUtils($this->{$this->name});
+
+        $this->loadComponent('CsvMigrations.CsvView');
     }
 
     /**
--- a/src/Events/BaseViewListener.php
+++ b/src/Events/BaseViewListener.php
@@ -45,6 +45,14 @@ abstract class BaseViewListener implements EventListenerInterface
         ['Documents', 'Files', 'Burzum/FileStorage.FileStorage']
     ];
 
+    /**
+     * Wrapper method that checks if Table instance has method 'findByLookupFields'
+     * and if it does, it calls it, passing along the required arguments.
+     *
+     * @param  \Cake\ORM\Entity  $entity Entity
+     * @param  \Cake\Event\Event $event  Event instance
+     * @return void
+     */
     protected function _lookupFields(Query $query, Event $event)
     {
         $methodName = 'findByLookupFields';
@@ -57,6 +65,14 @@ abstract class BaseViewListener implements EventListenerInterface
         $table->{$methodName}($query, $id);
     }
 
+    /**
+     * Wrapper method that checks if Table instance has method 'setAssociatedByLookupFields'
+     * and if it does, it calls it, passing along the required arguments.
+     *
+     * @param  \Cake\ORM\Entity  $entity Entity
+     * @param  \Cake\Event\Event $event  Event instance
+     * @return void
+     */
     protected function _associatedByLookupFields(Entity $entity, Event $event)
     {
         $methodName = 'setAssociatedByLookupFields';
@@ -68,6 +84,13 @@ abstract class BaseViewListener implements EventListenerInterface
         $table->{$methodName}($entity);
     }
 
+    /**
+     * Method that fetches action fields from the corresponding csv file.
+     *
+     * @param  \Cake\Network\Request $request Request object
+     * @param  string                $action  Action name
+     * @return array
+     */
     protected function _getActionFields(Request $request, $action = null)
     {
         $result = [];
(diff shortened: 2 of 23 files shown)


src/Template/Element/View/add.ctp
```diff
@@ -30,6 +30,12 @@ $formOptions = [
         'controller' => $this->request->controller,
         'action' => $this->request->action
     ],
+    'data-panels_url' => $this->Url->build([
+        'prefix' => 'api',
+        'plugin' => $this->request->plugin,
+        'controller' => $this->request->controller,
+        'action' => 'panels'
+    ]),
     'name' => Inflector::dasherize($moduleAlias),
```
flangofas: @georgeconstantinou This commit is a fix for task  task #2428, the form name is not needed anymore, so we can remove it.

src/Template/Element/View/add.ctp
```diff
@@ -30,6 +30,12 @@ $formOptions = [
         'controller' => $this->request->controller,
         'action' => $this->request->action
     ],
+    'data-panels_url' => $this->Url->build([
```
mamchenkov: @georgeconstantinou what's with the naming convention here? :)   dashes AND underscores? :)

georgeconstantinou: @mamchenkov a56b802

src/Controller/AppController.php
```diff
@@ -48,7 +51,6 @@ public function beforeFilter(\Cake\Event\Event $event)
      */
     public function index()
     {
-        $this->set('entities', $this->paginate($this->{$this->name}));
         $this->render('CsvMigrations.Common/index');
         $this->set('_serialize', ['entities']);
```
mamchenkov: @georgeconstantinou are entities still needed here?

src/Events/BaseViewListener.php
```diff
@@ -0,0 +1,224 @@
...
+
+        $controller = $request->controller;
+
+        if (is_null($action)) {
+            $action = $request->action;
+        }
+
+        $path = Configure::read('CsvMigrations.views.path') . $controller . DS . $action . '.csv';
+
+        if (is_readable($path)) {
```
mamchenkov: @georgeconstantinou Parser does the checks for file existence and readability, etc.  It either parses the file and returns the results or throws exceptions.  It's better therefor to use a `try..catch` block here instead of this `if`.  If there were no caught exceptions, then you can replace the result with the one returned from the parser.  Otherwise return an empty result (and, maybe, later, log the failure).

src/Events/BaseViewListener.php
```diff
@@ -0,0 +1,224 @@
...
+            return $result;
+        }
+
+        $result = $this->_containAssociations(
+            $associations,
+            $this->_nestedAssociations
+        );
+
+        // always include file associations
+        $result = array_merge(
```
mamchenkov: @georgeconstantinou if this is a recursive structure, then you might need `array_merge_recursive()`, as per this article:

http://jontai.me/blog/2011/12/array_merge_recursive-vs-array_replace_recursive/

src/Events/IndexViewListener.php
```diff
@@ -0,0 +1,231 @@
...
+
+    /**
+     * Pretty format identifier
+     */
+    const FORMAT_PRETTY = 'pretty';
+
+    /**
+     * Datatables format identifier
+     */
+    const FORMAT_DATATABLES = 'datatables';
```
mamchenkov: @georgeconstantinou you are already defining this in the `BaseViewListener`. It might not be needed here.

src/Events/IndexViewListener.php
```diff
@@ -0,0 +1,231 @@
...
+    /**
+     *
+     *
+     * @param  \Cake\ORM\Query   $query Query object
+     * @param  \Cake\Event\Event $event The event
+     * @return void
+     */
+    protected function _viewFields(Query $query, Event $event)
+    {
+        if (!in_array($event->subject()->request->query('format'), [static::FORMAT_DATATABLES])) {
```
mamchenkov: @georgeconstantinou it's not obvious why you are using `in_array()` here for string comparison.  Maybe add a comment like "Have a look at _prettifyEntities() on why in_array() is used".

src/Events/IndexViewListener.php
```diff
@@ -0,0 +1,231 @@
...
+        }
+
+        $conditions = [];
+        $tableName = $event->subject()->name;
+        foreach ($event->subject()->request->query('conditions') as $k => $v) {
+            if (false === strpos($k, '.')) {
+                $k = $tableName . '.' . $k;
+            }
+
+            $conditions[$k] = $v;
```
mamchenkov: @georgeconstantinou technically, there might be more than one condition for the same field.  For example, `amount > 0  AND amount > 100`.  So when building a list of conditions it's always safer to add them to `$conditions['AND']` array.  This PR is big enough already, so it's not necessary to do it here, but a comment, I think, is in order.

src/Controller/AppController.php
georgeconstantinou: eea53505a231b558388264d7700157822b702347

src/Events/BaseViewListener.php
georgeconstantinou: @mamchenkov 4b7d14b

src/Events/IndexViewListener.php
georgeconstantinou: @mamchenkov 08407b9

src/Events/IndexViewListener.php
georgeconstantinou: @mamchenkov eb02c14

PR #196, opened 2016-10-13
Remove loading of external libraries (task #2745)

georgeconstantinou:
These libraries have been moved now to the App level and will be loaded from there.


leonidsl:
:+1:

PR #199, opened 2016-10-14
Fix broken file associations (task #2699)

georgeconstantinou:
There is an issue with associated files when fetching records through the API. The problem started once we introduced the soft delete functionality. For the soft delete we are using [this](https://github.com/UseMuffin/Trash) plugin. It turned out that enabling the Trash behavior in our Tables it affects find queries with contained BelongsTo associations.

To overcome this issue we are currently fetching the associated record files using a more direct (hardcoded) method. Once the issue is resolved these commits can be reverted.

### DIFF
--- a/src/Events/BaseViewListener.php
+++ b/src/Events/BaseViewListener.php
@@ -10,6 +10,7 @@ use Cake\Network\Request;
 use Cake\ORM\AssociationCollection;
 use Cake\ORM\Entity;
 use Cake\ORM\Query;
+use Cake\Utility\Inflector;
 use CsvMigrations\FieldHandlers\CsvField;
 use CsvMigrations\FieldHandlers\FieldHandlerFactory;
 use CsvMigrations\Parser\Csv\MigrationParser;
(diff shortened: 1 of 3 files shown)


src/Events/BaseViewListener.php
```diff
@@ -177,6 +178,81 @@ protected function _databaseFields(array $fields, Event $event)
...
+     * @todo   this method is very hardcoded and has been added because of an issue with the soft delete
+     *         plugin (https://github.com/UseMuffin/Trash), which affects contain() functionality with
+     *         belongsTo associations. Once the issue is resolved this method can be removed.
+     */
+    protected function _includeFiles(Entity $entity, Event $event)
+    {
+        $associations = $event->subject()->{$event->subject()->name}->associations();
+
+        foreach ($associations as $docAssoc) {
+            if ('Documents' !== $docAssoc->className()) {
```
mamchenkov: @georgeconstantinou this probably shouldn't be hardcoded ... but given that this whole change shouldn't be there, it's fine :)

PR #211, opened 2016-10-21
Skip associated module records without index.csv defined (task #2851)

georgeconstantinou:


georgeconstantinou:
@mamchenkov I run the tests locally since Travis seems to be down.

PR #236, opened 2016-11-09
Integration tests

georgeconstantinou:


tests/Fixture/ArticlesFixture.php
```diff
@@ -0,0 +1,41 @@
...
+        'trashed' => ['type' => 'datetime', 'null' => true],
+        '_constraints' => [
+            'primary' => ['type' => 'primary', 'columns' => ['id']],
+            'unique' => ['type' => 'unique', 'columns' => ['name', 'id']]
+        ]
+    ];
+
+    public $records = [
+        [
+            'id' => 'd8c3ba90-c418-4e58-8cb6-b65c9095a2dc',
```
mamchenkov: @georgeconstantinou please use the sequential IDs like `000000-0000...0001`.  They are much easier on the eye and will prevent accidental errors with checking for the wrong ID.

georgeconstantinou: @mamchenkov 90803c2

PR #238, opened 2016-11-10
moved tabbed associations to the events (task #3007)

anvyst:
Decreased the size of the CsvViewComponent, to load most of the associations functionality into the Events for further refactoring and splitting it to other Events.

This pull request also introduces association labels, that let the user customize the tab names.

Example of associations panel:
```ini
[associations]
hide_associations = Foobar
association_labels[] = RelationIdFoobar,'Related Table 1'
association_labels[] = OtherFieldIdFoobar,'Managed Table 2'
```

### DIFF
--- a/src/ConfigurationTrait.php
+++ b/src/ConfigurationTrait.php
@@ -92,7 +92,6 @@ trait ConfigurationTrait
         $path = $pathFinder->find(Inflector::camelize($tableName));
         $parser = new Parser();
         $this->_config = $parser->parseFromPath($path);
-
         // display field from configuration file
         if (isset($this->_config['table']['display_field']) && method_exists($this, 'displayField')) {
             $this->displayField($this->_config['table']['display_field']);
@@ -122,8 +121,8 @@ trait ConfigurationTrait
             $this->hiddenAssociations($this->_config['associations']['hide_associations']);
         }
 
-        if (isset($this->_config['associations']['association_labels'])) {
-            $this->associationLabels($this->_config['associations']['association_labels']);
+        if (isset($this->_config['associationLabels'])) {
+            $this->associationLabels($this->_config['associationLabels']);
         }
 
         if (isset($this->_config['parent']['module'])) {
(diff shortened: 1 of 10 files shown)


mamchenkov:
@anvyst I'm a bit worried about the array values in the INI file and their parsing... Need to test :)

anvyst:
@mamchenkov , Did the tests for the association labels and covered pretty much all the basic data we can have in getConfig() in ConfigurationTrait.php (check expanded Foo/config.ini).

tests/data/CsvMigrations/migrations/Foo/config.ini
```diff
@@ -1,4 +1,18 @@
...
+[virtualFields]
+name = full_name
+
+[parent]
+module = TestModule
+redirect = self
+
+[associations]
+hide_associations = TestTable
+
```
mamchenkov: @anvyst please remove the empty line here .. to avoid potential misinterpretting of the following lines as outside of the section.

tests/data/CsvMigrations/migrations/Foo/array_in_config.ini
```diff
@@ -0,0 +1,11 @@
...
+alias = Foobar
+searchable = true
+display_field = name
+
+[associations]
+; Associations might be generated based
+; on registryAlias, that's not really
+; readable.
+association_labels[] = BarIdBaz,'Baz'
+association_labels[] = FieldIdTableName,'Table Name'
```
mamchenkov: @anvyst I'm still worried about the arrays :)  maybe better to switch it to something like:

```
[associations]
...

[associationLabels]
BarIdBas=Baz
FieldIdTable='Table Name'
```

anvyst: @mamchenkov , moved the stuff into separate section of config.ini

src/ConfigurationTrait.php
```diff
@@ -98,6 +117,18 @@ protected function _setConfiguration($tableName)
...
+        if (isset($this->_config['associations']['hide_associations'])) {
+            $this->hiddenAssociations($this->_config['associations']['hide_associations']);
+        }
+
+        if (isset($this->_config['associationLabels'])) {
+            $this->associationLabels($this->_config['associationLabels']);
+        }
+
+        if (isset($this->_config['parent']['module'])) {
+            $this->parentField($this->_config['parent']['module']);
```
mamchenkov: @anvyst I think there is a problem here ... in my understanding, the configuration should look like this:

```
[parent]
module=Examples
field=example_id
redirect=self
```

If that is true, then you are setting the `$this->_parentField` to the module name, not the field name.

PR #249, opened 2016-11-15
vCal support (task #2824)

anvyst:
vCal support can be enabled for Reminder fields.
The user should add extra line in config.ini of the related table, to get reminders:
```ini
; <Model>/config.ini
[table]
allow_reminders = Users
```
Upon saving/editing the record afterSave hook will be initiated. The script will trigger the lookup for the associated record for the `Users` table (or others specified in the `allow_reminders` comma separated line), and pick up `email` record from the entity.

The email will containt `event.ics` file with the Event information. The format is applicable for Office 365 and Gmail.

Timezone is used from `APP/config/bootstrap.php`:
```php
date_default_timezone_set('Europe/Athens'); //using native IANA standard
```

Reference: [DST time changes worldwide](https://www.timeanddate.com/time/dst/2016.html)


src/Events/ModelAfterSaveListener.php
```diff
@@ -0,0 +1,201 @@
...
+            //initial reminder event
+            $vEvent->setAttendees($vAttendees)
+                ->setOrganizer($vOrganizer)
+                ->setSummary($entity->subject);
+
+            $endDate = new Time($entity->start_date->format('Y-m-d H:i:s'));
+            $endDate->modify("+ {$durationMinutes} minutes");
+
+            //setting the rule on switching hours for Daylight Saving Time
+            $vTimezoneRuleDst = new \Eluceo\iCal\Component\TimezoneRule(\Eluceo\iCal\Component\TimezoneRule::TYPE_DAYLIGHT);
```
mamchenkov: @anvyst this method seems to be doing too many things.  Can you please break it up into a few protected methods or something along those lines.  Ideally, this method should provide a higher level overview of what is going on, with implementation details being abstracted into their own methods.  Something like (pseudo-code):

```php
public function sendCalendarReminder() {
  $config = $this->getConfiguration();
  if (!$config->allowReminders) {
    return;
  }

  $attendees = $this->getAttendees();
  if (empty($attendees)) {
    return;
  }

  foreach ($attendees as $attendee) {
    $event = $this->getEvent();
    $event->addAttendees($attendees);
    $calendar = $this->getCalendar();
    $calendar->addEvent($event);
    $email->createEmail($calendar);
    $email->send($attendee);
}
```

PR #253, opened 2016-11-22
Re-arrangement of tabs into events (Task #2995)

anvyst:
## Warning

This patch might break the compatibility of previous plugin versions (in the matter of being customly used in other projects). You should use associations settings in `config.ini` to hide unwanted tabs and make sure your `associations` remain consistent, example:

```ini
[associations]
; in case of ordered_by is the manyToOne association to Users table.
hide_associations=OrderedByUsers,OtherIdTableName

[associationLabels]
FooIdUsers = 'Related Users'
AnotherIdDocuments = 'Document Relation'
```
### Changes

Following patch introduces evented tabs.
It mostly relies on `associations` of the ORM\Table|Entity we operate with.

* In the `view.ctp` we pass the entity into `CsvMigrations.View.View.getTabList` event and iterate through the associations. Event result holds the list of tabs with its corresponding labels, and required class source/target information on the association.

* Each tab represents an event `getTabContent` that has also `after|before` events respectively, to allow loading things within the tab-panel before and after actual tab content.

* In case of non-manyToMany relation 'link|unlink' button is hidden. Typeahead field that allows us to add & link manyToMany records is only visible in the tabs with manyToMany association type.

* Now deleting `documents` also checks for all the mentioned UUID in the related records, and sets them to `NULL` if any found. **It prevents breaking of FieldHandlers to break** while trying to associate non-empty related/has_many record while rendering the input field.

* Modal window `modal-id|data-target` is now supporting "more" unique id's for not only manyToMany associations. It uses aliases that CakePHP establishes on `FieldIdTableName` in the core.


anvyst:
@mamchenkov , that's why I've sent a pull request to `develop` branch, that shouldn't affect current master.

PR #266, opened 2016-11-30
Adding alias titles for pages (task #2520)

anvyst:
If we're using Qobo-admin-panel plugin for rendering view, we can overwrite the title, for using table aliases, that's more friendly for the users.

### DIFF
--- a/src/Events/LayoutListener.php
+++ b/src/Events/LayoutListener.php
@@ -1,0 +1,39 @@
+<?php
+namespace CsvMigrations\Events;
+
+use Cake\Event\Event;
+use Cake\Event\EventListenerInterface;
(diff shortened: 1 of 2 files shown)


src/Events/LayoutListener.php
```diff
@@ -0,0 +1,39 @@
...
+    public function implementedEvents()
+    {
+        return [
+            'QoboAdminPanel.Layout.Head' => 'getMyHead'
+
+        ];
+    }
+
+    /**
+     * getAllReports method
```
mamchenkov: @anvyst reports? :)

anvyst: @mamchenkov , as this pull request was closed, i'll cross-reference the patch here #267

PR #268, opened 2016-12-01
Fixing tab events (task #3142)

anvyst:
The issue was caused by incorrectly handling the data of the event.
* Event data passed/returned was standardized, to avoid collisions within TabContentCell
* Using `display` default method with custom templates passed, gets broken with camel/snake cased custom template names, thus had to simplify template name.

### DIFF
--- a/src/View/Cell/TabContentCell.php
+++ b/src/View/Cell/TabContentCell.php
@@ -23,7 +23,7 @@ class TabContentCell extends Cell
      * List of accessible displayTemplates for TabContent
      * @var array
      */
-    protected $_displayTemplates = ['display', 'panelTable'];
+    protected $_displayTemplates = ['display', 'panel'];
 
 
     /**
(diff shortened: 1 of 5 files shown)


src/View/Cell/TabContentCell.php
```diff
@@ -67,17 +67,15 @@ public function display(array $data)
     {
         $this->template = $this->_getDisplayTemplate($data['content']);
```
anvyst: @mamchenkov , pushed the patch, awaiting for the tests.

src/View/Cell/TabContentCell.php
```diff
@@ -67,17 +67,15 @@ public function display(array $data)
...
-                $data['content']['length'] = $data['content']['records']->count();
-                $data['content']['records'] = $data['content']['records']->toArray();
-            } elseif ($data['content']['records'] instanceof Entity) {
-                $tmp[] = $data['content']['records'];
-                $data['content']['records'] = $tmp;
-                $data['content']['length'] = count($data['content']['records']);
-            } elseif (is_array($data['content'])) {
-                $data['content']['length'] = count($data['content']['records']);
-            }
+        if ($data['content']['records'] instanceof ResultSet) {
```
mamchenkov: @anvyst please check that the `$data['content']['records']` is set/exists/etc.

PR #291, opened 2016-12-13
Images filehandler 3139

anvyst:
* *ImagesFieldHandler* is introduced to handle thumbnails and image store in Burzum plugin
* *BaseFileFIeldHandler* got refactored, specific code got moved into children classes (files, images field handlers).

### DIFF
--- a/src/FieldHandlers/BaseFileFieldHandler.php
+++ b/src/FieldHandlers/BaseFileFieldHandler.php
@@ -158,24 +158,6 @@ class BaseFileFieldHandler extends RelatedFieldHandler
     /**
      * {@inheritDoc}
      */
-    public function renderValue($table, $field, $data, array $options = [])
-    {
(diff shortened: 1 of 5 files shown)


src/FieldHandlers/BaseFileFieldHandler.php
```diff
@@ -161,20 +161,15 @@ public function renderInput($table, $field, $data = '', array $options = [])
     public function renderValue($table, $field, $data, array $options = [])
```
mamchenkov: @anvyst I think this method does do much anymore :)   There are a few IFs, which are leftovers of the old code.  Can you please check it.

anvyst: @mamchenkov , I've removing this method call, as non of the children classes is using it, and it doesn't carry any logic for the children classes.

PR #297, opened 2016-12-16
Fixing embedded file uploads (task #3195)

anvyst:
Embedded inputs on the forms were broken due to recent new File upload functionality introduced.

### DIFF
--- a/src/FieldHandlers/FilesFieldHandler.php
+++ b/src/FieldHandlers/FilesFieldHandler.php
@@ -46,8 +46,13 @@ class FilesFieldHandler extends BaseFileFieldHandler
      */
     protected function _renderInputWithoutData($table, $field, $options)
     {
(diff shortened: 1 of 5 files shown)


src/FieldHandlers/FilesFieldHandler.php
```diff
@@ -46,8 +46,13 @@ public function renderInput($table, $field, $data = '', array $options = [])
      */
     protected function _renderInputWithoutData($table, $field, $options)
     {
+        $fieldName = $this->_getFieldName($table, $field, $options);
+        if (!empty($options['embedded'])) {
+            $fieldName = $table->alias() . '.' . $field;
+        }
```
georgeconstantinou: @anvyst I think you can avoid this `if` check by not passing the `$options` variable to `_getFieldName()` method.

PR #300, opened 2016-12-20
Field handler fixes

mamchenkov:
A variety of small fixes around the field handlers.  Most noticeable:

* Improve the handling of the `$data` parameter in field handlers (figure things out from the type of the variable).
* Improve the handling of the `$options['entity']` parameter, especially when it is not an Entity instance.
* Improve display value in Related Field Handler, including the API and UI, like the select drop-downs.  The lookup is now recursive and is done properly through the field handlers, bringing more consistency.


georgeconstantinou:
Excellent work @mamchenkov  :tada:

mamchenkov:
Thanks @georgeconstantinou ! Wouldn't be possible without you :)

PR #328, opened 2017-01-21
Field handlers refactoring (task #3317)

mamchenkov:
This PR is here for quick access to code review and test runs.  It's work in progress.

Here are a few things that have been done so far:

1. Major reorganization of the field handler classes tree hierarchy.  More base classes were introduced in order make reusing common functionality easier, and to decrease an average size of each class.
2. A variety of search-related methods were removed from the interface and grouped into a single `getSearchOptions()` method.  This will require changes outside the plugin as well.  In particular, some search-related events used those methods in project-template-cakephp.
3. `FileFieldHandler` was removed.
4. `DefaultFieldHandler` was removed, as we now have to know exactly which field we are loading and for which table.
5. `renderName()` method was introduced for consistent rendering of field names.  This will also be much easier to extend soon, with per-field configurations like labels, default values, etc.
6. `FieldHandlerFactory` has been reviewed and improved, especially to easy a near-future simplification of that class to the classical Factory pattern.
7. Path finders were slightly improved.
8. Interoperability with non-CSV modules was improved, particularly via the `getFieldsDefinitions()` method, which now supports stub fields.
9. `ListTrait` was removed, with all the relevant code falling into the lists field handlers.
10. `CsvField` and `DbField` class were revised to provide additional flexibility and stability.
11. Leftovers of the BootstrapUI and QoboAdminPanel were removed, specifically to avoid test failures with the newest CakePHP v3.3.13.
12. Quite a few new tests were added, pushing the code coverage up by almost 10%.
13. A few coding style and design fixes were applied all over the place to make the code more maintainable.
14. A lot of documentation improvements were made all around.  In particular, we are now trying to avoid `{inheritDoc}` tags, as not all tools support them, and it's not obvious for the human to figure out from where the documentation should be obtained (lots of class inheritance, traits, external sources, etc).  Instead, we should copy-paste the documentation to have it right next to the method/variable/constant.  The ultimate source of truth should always be interface documentation.
15. Added `sanitizeValue()` to the interface and to all the classes in the hierarchy.  The default behavior down in the `BaseFieldHandler` is to not modify the value, as we don't have enough understanding of the consequences yet.  But each branch of the hierarchy tree and each class by itself can introduce appropriate sanitization, like we already do for strings, numbers, URLs, etc.

Regarding the changes in the field handler classes hierarchy, consider the two diagrams below, illustrating the old tree and the current one.

**BEFORE**: The old class tree of field handlers

old

**AFTER**:  The new class tree of field handlers (so far)

new

Here are a few things that are still pending:

- [ ] Make lists, related, and file handlers smaller and easier.  They are still too bulky and unreadable.
- [ ] Upgrade the rest of the system that uses field handlers - views and APIs in particular.
- [ ] Increase the test coverage to check for at least the most basic functionality (much like `fieldToDb()` is tested now).
- [ ] Have a quick look at performance.  Firstly, are there any unintentional changes, introduced by the new structure.  Secondly, make the whole thing faster, especially when it comes to lots of relationships and lots of files.


src/MigrationTrait.php
```diff
@@ -28,32 +28,70 @@
...
+                throw new RuntimeException("Failed getting field definitions for unknown module");
+            }
+
+            $pathFinder = new MigrationPathFinder;
+            $path = $pathFinder->find($moduleName);
+
+            // Parser knows how to make sure that the file exists.  But it can
+            // also throw other exceptions, which we don't want to avoid for
+            // now.
+            if (is_readable($path)) {
```
georgeconstantinou: @mamchenkov we need to check how this method will behave if `$path` is not readable.

src/MigrationTrait.php
```diff
@@ -86,7 +92,7 @@ public function getFieldsDefinitions(array $stubFields = [])
 
         // Merge $result with $stubFields
         foreach ($stubFields as $field => $definition) {
-            if (!in_array($field, array_keys($result))) {
+            if (empty($result) || !in_array($field, array_keys($result))) {
```
georgeconstantinou: @mamchenkov if `$result` is not empty the stub-field won't be added to it. Is this the expected behavior?

src/FieldHandlers/FieldHandlerFactory.php
```diff
@@ -214,21 +217,21 @@ protected function _getTableInstance($table)
      * @param  string|array  $field Field name
      * @return object               FieldHandler instance
      */
-    protected function _getHandler($table, $field)
+    protected function _getHandler(Table $table, $field)
     {
-        if (!is_object($table)) {
-            throw new \InvalidArgumentException("Table parameter is not an instance of Table class");
+        if (empty($table)) {
```
georgeconstantinou: @mamchenkov since there is a class type hint in the function parameter I don't think this check is required.

src/FieldHandlers/BaseFieldHandler.php
```diff
@@ -227,8 +231,11 @@ public function renderInput($data = '', array $options = [])
         $options = array_merge($this->defaultOptions, $this->fixOptions($options));
         $data = $this->_getFieldValueFromData($data);
 
-        return $this->cakeView->Form->input($this->_getFieldName($options), [
+        $fieldName = $this->table->alias() . '.' . $this->field;
```
georgeconstantinou: @mamchenkov for all of these cases we can switch to `$this->table->aliasField($this->field)`

src/FieldHandlers/BaseFileFieldHandler.php
```diff
@@ -102,11 +102,11 @@ public function renderInput($data = '', array $options = [])
         $data = $this->_getFieldValueFromData($data);
         $relatedProperties = $this->_getRelatedProperties($options['fieldDefinitions']->getLimit(), $data);
 
-        $fieldName = $this->_getFieldName($options);
+        $fieldName = $this->table->alias() . '.' . $this->field;
 
         $input['html'] = '';
         $input['html'] .= '<div class="form-group' . ((bool)$options['fieldDefinitions']->getRequired() ? ' required' : '') . '">';
-        $input['html'] .= $this->cakeView->Form->label($this->field);
+        $input['html'] .= $options['label'] ? $this->cakeView->Form->label($this->field, $options['label']) : '';
```
georgeconstantinou: @mamchenkov since we are merging with `defaultOptions`  'label' parameter will not always be available?

mamchenkov: @georgeconstantinou this for the case of disabling labels (for example in combined fields).  So you can pass `$options['label'] = false`.

src/MigrationTrait.php
mamchenkov: @georgeconstantinou good catch! Thanks

src/FieldHandlers/FieldHandlerFactory.php
mamchenkov: @georgeconstantinou :+1:

src/FieldHandlers/BaseFieldHandler.php
mamchenkov: @georgeconstantinou thanks! Didn't know about that one.

mamchenkov:
I have a strong suspicion that the error that breaks the test run is related to the release of CakePHP v3.3.13 :)

https://bakery.cakephp.org/2017/01/28/cakephp_3313_released.html

PR #333, opened 2017-01-31
Lookup results sorting (task #3285)

georgeconstantinou:
Sorting of lookup results is now applied on the actual query before is run. The sorting is applied by the module's typeahead fields, and if there is a parent module then its display field takes precedence over the module's typeahead fields.

For example, if a module has typeahead fields `['name', 'type']` then these will be used for sorting the results. If it has a parent module with display field 'title', then the sorting order will change to `['title', 'name', 'type']`.

### DIFF
--- a/src/Events/LookupListener.php
+++ b/src/Events/LookupListener.php
@@ -36,26 +36,24 @@ class LookupListener extends BaseViewListener
     public function beforeLookup(Event $event, Query $query)
     {
         $request = $event->subject()->request;
-        if (!empty($request->query['query'])) {
-            $table = $event->subject()->{$event->subject()->name};
+        $table = $event->subject()->{$event->subject()->name};
 
-            $typeaheadFields = [];
-            // Get typeahead fields from configuration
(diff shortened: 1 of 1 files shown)


src/Events/LookupListener.php
```diff
@@ -36,26 +36,24 @@ public function implementedEvents()
...
 
-            if (!empty($typeaheadFields)) {
-                foreach ($typeaheadFields as $field) {
-                    $query->orWhere([$field . ' LIKE' => '%' . $request->query['query'] . '%']);
-                }
-            } else {
-                throw new RuntimeException("No typeahead or display field configured for " . $table->alias());
-            }
+        if (empty($fields)) {
+            throw new RuntimeException("No typeahead or display field configured for " . $table->alias());
```
mamchenkov: @georgeconstantinou I think it's better to have a random order or order by some known field like id or created or modified.  Wrong sorting is annoying, but not as annoying as an error 500.

PR #346, opened 2017-02-07
Migration issue with fieldsToDb

anvyst:
When the migration is first hit by `cake migrations migrate`, method `createFromCsv` method
tries to instantiate TableRegistry of the table that hasn't been created.

We should instantiate the FieldHandler object from the Factory that stores the settings on the specific field type, and convert it to Db field without instantiating the table.

Current proof of concept does create a table but throws a lot of exceptions on the `fieldToDb` method:

```
SQLSTATE[42S02]: Base table or view not found: 1146 Table 'db.tablename' doesn't exist
SQLSTATE[42S02]: Base table or view not found: 1146 Table 'db.tablename' doesn't exist
```

### DIFF
--- a/src/CsvMigration.php
+++ b/src/CsvMigration.php
@@ -64,9 +64,9 @@ class CsvMigration extends AbstractMigration
     public function csv(Table $table, $path = '')
     {
         $this->_fhf = new FieldHandlerFactory();
+
         $this->_table = $table;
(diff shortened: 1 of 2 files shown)


mamchenkov:
Closing.  Fixing via PR #347 .

But thanks anyway :)

PR #351, opened 2017-02-10
Fields ini support (task #3316)

georgeconstantinou:
Added support for parsing and using `fields.ini` configuration. Currently, supports defining custom label and default value for each of the module fields.

### DIFF
--- a/src/FieldHandlers/BaseFieldHandler.php
+++ b/src/FieldHandlers/BaseFieldHandler.php
@@ -597,16 +597,14 @@ abstract class BaseFieldHandler implements FieldHandlerInterface
     /**
      * Parse and return fields ini configuration file.
      *
-     * @return array
+     * @return string
      */
-    protected function _parseFieldsIni()
+    protected function _getFieldsIniPath()
     {
-        $result = [];
+        $result = '';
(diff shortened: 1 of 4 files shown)


src/FieldHandlers/BaseFieldHandler.php
```diff
@@ -513,6 +527,90 @@ protected function _getFieldValueFromData($data, $field = null)
...
+     * are requested then the full field ini configuration is returned.
+     *
+     * If no field ini configration was found then the return value will be an empty
+     * array if multiple parameters were requested or null if one parameter was
+     * requested.
+     *
+     * @param string|array|null $params Field ini parameters
+     * @return string|array|null
+     */
+    public function getFieldsIniParams($params = null)
```
mamchenkov: @georgeconstantinou I think maybe the parser is the better place for it ... what do you think?

georgeconstantinou: @mamchenkov yes great idea

georgeconstantinou: @mamchenkov done d60471a

PR #354, opened 2017-02-14
Bake CSV Module task (task #3349)

georgeconstantinou:
This shell task automatically bakes app and api controllers, model table class, model entity class and generates basic CSV configuration files (such as `migrations.csv`, `config.ini`, `fields.ini`,
`add.csv`, `edit.csv` etc).

You can run this task using:

```
bin/cake bake csv_module Articles
```


mamchenkov:
Beautiful work, @georgeconstantinou ! :)

PR #368, opened 2017-03-07
fields.ini format parameter (task #2970)

georgeconstantinou:
Add check for `format` parameter in `fields.ini` configuration. If the parameter value is `raw`, then we skip formatting the number values.


src/FieldHandlers/BaseNumberFieldHandler.php
```diff
@@ -77,6 +78,13 @@ protected function formatValue($data, array $options = [])
     {
         $result = (float)$data;
 
+        $path = $this->_getFieldsIniPath();
```
mamchenkov: @georgeconstantinou this should probably be another path finder.

PR #371, opened 2017-03-08
Auto increment Event listener (task #3481)

georgeconstantinou:
Added event listener for handling fields defined as auto-incremental using `fields.ini` configuration. Currently, supported paramaters are `auto-increment` and `min`. Currently stepping is always one,
so an integer starting from 0 will increment like 1, 2, 3 etc and a decimal starting from 0.5 will increment like 1.5, 2.5, 3.5 etc.

To define a field as auto-incremental you just add the following lines to `fields.ini` configuration file:

```ini
[field_name_goes_here]
auto-increment = true
```

You can optionally set the `min` parameter to specify the starting point of the auto-increment functionality. See example below:

```ini
[field_name_goes_here]
auto-increment = true
min = 100
```

So the first record created, with the above configuration in place, will have the field value set to 100. The next one will be set to 101 etc.

### DIFF
--- a/src/Events/Model/AutoIncrementEventListener.php
+++ b/src/Events/Model/AutoIncrementEventListener.php
@@ -63,12 +63,13 @@ class AutoIncrementEventListener implements EventListenerInterface
             $query = $event->subject()->find('withTrashed');
             $query->select([$field => $query->func()->max($field)]);
             $max = $query->first()->toArray();
+            $max = (float)$max[$field];
 
             if (empty($options['min'])) {
-                $entity->{$field} = $max[$field] + 1;
+                $entity->{$field} = $max + 1;
             } else {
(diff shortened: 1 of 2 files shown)


src/Events/Model/AutoIncrementEventListener.php
```diff
@@ -0,0 +1,121 @@
...
+            }
+
+            return;
+        }
+
+        foreach ($autoIncrementFields as $field => $options) {
+            // get max value
+            $query = $event->subject()->find('withTrashed');
+            $query->select([$field => $query->func()->max($field)]);
+            $max = $query->first()->toArray();
```
mamchenkov: @georgeconstantinou we need to test this for empty table and 0 ...

PR #374, opened 2017-03-10
Extra classes on select elements for FieldHandlers (task #3371)

anvyst:
In order to hook any JS callbacks it's much easier to do it via custom classes rather then input names and data-attributes that might dynamically change.

### DIFF
--- a/src/Template/Element/FieldHandlers/ListFieldHandler/input.ctp
+++ b/src/Template/Element/FieldHandlers/ListFieldHandler/input.ctp
@@ -1,7 +1,7 @@
 <div class="form-group<?= $required ? ' required' : '' ?>">
     <?= $label ? $this->Form->label($name, $label) : ''; ?>
     <?= $this->Form->select($name, $options, [
-        'class' => 'form-control',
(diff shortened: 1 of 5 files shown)


anvyst:
@mamchenkov , extraClasses options added to BaseFieldHandler so we can use it in other field handlers as well.

src/Template/Element/FieldHandlers/ListFieldHandler/input.ctp
```diff
@@ -1,8 +1,8 @@
 <div class="form-group<?= $required ? ' required' : '' ?>">
     <?= $label ? $this->Form->label($name, $label) : ''; ?>
     <?= $this->Form->select($name, $options, [
-        'class' => 'form-control',
+        'class' => 'form-control' . ( (isset($extraClasses) && !empty($extraClasses)) ? ' ' . $extraClasses : null ),
```
mamchenkov: @anvyst I think this line is also needed on the other templates.

anvyst: @mamchenkov , added to those that mention 'class' parameter in the options array.

PR #404, opened 2017-05-11
Add index to all related fields (task #3803)

georgeconstantinou:


mamchenkov:
@georgeconstantinou beautiful!

PR #405, opened 2017-05-12
Thumbnails via API (task #3808)

georgeconstantinou:
Attaches thumbnail paths into the file storage entity.

### DIFF
--- a/src/FileUploadsUtils.php
+++ b/src/FileUploadsUtils.php
@@ -63,7 +63,7 @@ class FileUploadsUtils
     /**
      * Contructor method
      *
-     * @param UploadTable $table Upload Table Instance
+     * @param \Cake\ORM\Table $table Upload Table Instance
      */
     public function __construct(UploadTable $table)
     {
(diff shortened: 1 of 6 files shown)


src/FileUploadsUtils.php
```diff
@@ -2,13 +2,17 @@
...
 
 use Burzum\FileStorage\Storage\StorageManager;
+use Cake\Core\App;
 use Cake\Core\Configure;
 use Cake\Event\Event;
 use Cake\Event\EventManager;
+use Cake\Log\Log;
 use Cake\ORM\Entity;
 use Cake\ORM\TableRegistry;
 use Cake\ORM\Table as UploadTable;
```
mamchenkov: @georgeconstantinou is there a reason for us aliasing the class name here?

georgeconstantinou: @mamchenkov not really, I guess it might have been used before the last refactoring of the uploads functionality.

PR #406, opened 2017-05-12
Field value renderers (task #3809)

mamchenkov:
This PR pushes some of the rendering functionality (in particular `renderValue()`) from within field handler classes into the renderer classes.  This change allows for the following benefits:

* Simpler, smaller, and more configurable field handlers.
* Simpler, smaller, and more configurable renderers.
* Greater support for a variety of options in the `fields.ini` file.
* Easier and simpler rendering availability for code that needs the same consistency, but does not have enough knowledge to use full-blown field handlers.
* Greater control over the sanitization of the rendered results.

In particular, the following improvements were done:

* Decimal precision can be now controlled on the numeric fields.
* Date and time format can be now controlled on the date/time/datetime fields.
* Link target (new window, self, etc) can be now controlled on the URL fields.
* List items can be now controlled in the CSV lists and similar fields.
* List name can be now controlled in the Dblist fields.
* Labels can be now controlled in the boolean fields (0/1, yes/no, on/off, and more).

Also, the new Link renderer has been implemented, which can be used to render field value as a hyperlink to an arbitrary URL, with a placeholder replaced by the field value.

renders


georgeconstantinou:
@mamchenkov excellent job :fireworks:

PR #410, opened 2017-05-26
[WIP] Bugfix combined fields value (task #3920)

georgeconstantinou:
Always use entity object when fetching value for combined fields.


src/FieldHandlers/BaseCombinedFieldHandler.php
```diff
@@ -42,7 +42,7 @@ public function renderInput($data = '', array $options = [])
             $options['label'] = false;
             $fieldName = $this->field . '_' . $suffix;
 
-            $fieldData = $this->_getFieldValueFromData($data, $fieldName);
```
mamchenkov: @georgeconstantinou we need to talk about this one.  If there's something wrong with `_getFieldValueFromData()` method, we need to look at it, as it might have implications on other things.

PR #422, opened 2017-06-28
Changed top entity and actions menu definitions - task #809

stepanov:
Changed menu definition to fix new menu render approach.


src/Template/Element/Menu/view_top.ctp
```diff
@@ -11,6 +11,8 @@ if (!empty($this->request->plugin)) {
 
 $menu = [];
 
+$newMenu = [];
```
georgeconstantinou: @stepanov is this variable used anywhere?

stepanov: @georgeconstantinou removed it

PR #426, opened 2017-07-17
Implement database seeders with fake data (task #2790)

christosapl:


src/Shell/SeedShell.php
```diff
@@ -0,0 +1,372 @@
...
+use Cake\ORM\TableRegistry;
+use CsvMigrations\MigrationTrait;
+use Faker\Factory;
+use Qobo\Utils\ModuleConfig\ModuleConfig;
+
+class SeedShell extends Shell
+{
+    use MigrationTrait;
+
+    protected $numberOfData = 1;
```
mamchenkov: @christosapl I think 'numberOfRecords' is a better name.

src/Shell/SeedShell.php
```diff
@@ -0,0 +1,372 @@
...
+    /**
+     * Get Module ids in an array.
+     *
+     * @param string $moduleName module name
+     * @return array
+     */
+    protected function getModuleIds($moduleName)
+    {
+        $table = TableRegistry::get($moduleName);
+        $query = $table->find()->select('id')->toArray();
```
georgeconstantinou: @christosapl don't rely on 'id' being the primary key. Just use `$table->getPrimaryKey()` to get the primary key of the table.

src/Shell/SeedShell.php
```diff
@@ -0,0 +1,372 @@
...
+            foreach ($module as $field) {
+                if (empty($field['type'])) {
+                    continue;
+                }
+                if (strpos($field['type'], 'related') !== false) {
+                    //get related module
+                    $type = $this->getStringEnclosedInParenthesis($field['type']);
+                    /*if ( in_array($type, $this->skipModules) ){
+                        continue;
+                    }*/
```
georgeconstantinou: @christosapl enable this code if needed, otherwise remove it.

src/Shell/SeedShell.php
mamchenkov: @christosapl you might also end up running out of memory, as some tables can have hundreds of thousands of records.  Either limit the resultset, or fetch a single random ID every time you need one.

src/Shell/SeedShell.php
```diff
@@ -0,0 +1,369 @@
...
+        }
+
+        return $csvFiles;
+    }
+
+    /**
+     * Insert data into module.
+     *
+     * @param string $moduleName module name.
+     */
```
georgeconstantinou: @christosapl missing `@return` parameter. You should set it to void if nothing is returned.

src/Shell/SeedShell.php
```diff
@@ -0,0 +1,369 @@
...
+        }
+
+        return $index;
+    }
+
+    /**
+     * Hierarchical insert data into modules (based on index hierarchy).
+     *
+     * @param array $index index.
+     */
```
georgeconstantinou: @christosapl same here, missing `@return` parameter.

src/Shell/SeedShell.php
```diff
@@ -0,0 +1,369 @@
...
+            $this->checkHierarchyForModule($moduleName, $index);
+        }
+    }
+
+    /**
+     * Check the hierarchy for each module recursively and populate data.
+     *
+     * @param string $moduleName module name.
+     * @param array $index index.
+     */
```
georgeconstantinou: @christosapl same here, missing `@return` parameter.

src/Shell/SeedShell.php
```diff
@@ -0,0 +1,369 @@
...
+    protected function checkHierarchyForModule($moduleName, array $index = [])
+    {
+        if (empty($moduleName)) {
+            return;
+        }
+
+        if (in_array($moduleName, $this->modulesPolpulatedWithData)) {
```
georgeconstantinou: @christosapl can you add a comment here, why you are skipping the logic?

src/Shell/SeedShell.php
```diff
@@ -0,0 +1,369 @@
...
+use Qobo\Utils\ModuleConfig\ModuleConfig;
+
+class SeedShell extends Shell
+{
+    use MigrationTrait;
+
+    protected $numberOfRecords = 1;
+    protected $modules = [];
+    protected $modulesPolpulatedWithData = [];
+    protected $skipModules = [];
```
georgeconstantinou: @christosapl can you add comments on all the class properties?

src/Shell/SeedShell.php
```diff
@@ -0,0 +1,369 @@
...
+    {
+        if (empty($moduleName)) {
+            return;
+        }
+
+        if (in_array($moduleName, $this->modulesPolpulatedWithData)) {
+            return;
+        }
+
+        if (!empty($index[$moduleName]) && is_array($index[$moduleName])) {
```
georgeconstantinou: @christosapl please add a comment here that you first generate fake data for modules without any relations.

Also, whats the expected behavior, if all modules are associated with one another?

src/Shell/SeedShell.php
```diff
@@ -0,0 +1,451 @@
...
+                $value = $faker->unique()->text(20);
+                break;
+            case 'text':
+                $value = $faker->unique()->paragraph();
+                break;
+            case 'phone':
+                $value = $faker->unique()->phoneNumber;
+                break;
+            case 'decimal':
+            case 'money':
```
mamchenkov: @christosapl `money` is a combined field, which is represented in the database by `fieldname_amount` and `fieldname_currency`.

src/Shell/SeedShell.php
```diff
@@ -0,0 +1,451 @@
...
+                $value = $faker->unique()->randomFloat();
+                break;
+            case 'integer':
+                $value = $faker->unique()->numberBetween();
+                break;
+            case 'image':
```
mamchenkov: @christosapl this one can probably be skipped, as we handle images and files differently (via the FileStorage relationship in the opposite direction - table, field, id or something).

src/Shell/SeedShell.php
```diff
@@ -0,0 +1,451 @@
...
+                $value = $faker->unique()->imageUrl();
+                break;
+            case 'email':
+                $value = $faker->unique()->email;
+                break;
+            case 'datetime':
+                $value = $faker->unique()->dateTime;
```
mamchenkov: @christosapl I think it's better to use a specific format ('yyyy-MM-dd HH:mm:ss' or whatever is the string for phinx).

src/Shell/SeedShell.php
```diff
@@ -0,0 +1,451 @@
...
+
+class SeedShell extends Shell
+{
+    use MigrationTrait;
+
+    /**
+     * Number of records to be added for each Module.
+     * @var int
+     */
+    protected $numberOfRecords = 5;
```
mamchenkov: @christosapl please also increase to 20 .. we show 10 records per page by default and this will allow us to have at least a couple of pages.

PR #448, opened 2017-08-09
[WIP] Code climate fixes

mamchenkov:


mamchenkov:
OK, I think, that's enough for now ... :)

Just for the record, here's what I've started with:

before

And here's how it looks now:

after

It's not perfect, but it's understandable (and more importantly, maintainable) now.

PR #470, opened 2017-09-15
Added current user ID to the event options

stepanov:


src/Table.php
```diff
@@ -55,6 +55,7 @@ class Table extends BaseTable
     public function setCurrentUser($user)
     {
         $this->_currentUser = $user;
+        error_log('Set current user: ' . print_r($this->_currentUser, true) . "\n", 3, '/tmp/sync.log');
```
mamchenkov: @stepanov please remove the error log.

src/Table.php
```diff
@@ -117,6 +117,9 @@ public function beforeSave(Event $event, EntityInterface $entity, ArrayObject $o
      */
     public function afterSave(Event $event, EntityInterface $entity, ArrayObject $options)
     {
+        $user = $this->getCurrentUser();
+        $options['current_user'] = $user['id'];
```
mamchenkov: @stepanov any particular reason for not pushing the whole user record?

stepanov: @mamchenkov I need ID to link record with user's account. Don't see the reason to pass whole record for that.

mamchenkov: @stepanov the `afterSave` event is not just for you :)  since we have the user information in the event, we might do other things with it (send notifications/emails/logs/etc).  I think we should send the whole record, since we are modifying the event options anyway.

stepanov: @mamchenkov changed to pass all user's data instead of just ID

PR #507, opened 2017-12-11
Proof of concept implementation for Providers (task #5061)

mamchenkov:
* Created Field Handler configurations.
* Created Field Handler providers.
* Refactored giant Field Handler classes and complex hierarchy into flat list of configs and providers.


georgeconstantinou:
@mamchenkov excellent work :tada:

PR #522, opened 2018-02-08
Replace DB_FIELD_TYPE constant with a method (task #5061)

mamchenkov:
Replace `DB_FIELD_TYPE` constant in field handler classes, with the call to `getDbFieldType()` public static method call in `BaseFieldHandler` class, and a set of `DbFieldType` providers.

### DIFF
--- a/src/FieldHandlers/BaseFieldHandler.php
+++ b/src/FieldHandlers/BaseFieldHandler.php
@@ -35,11 +35,6 @@ use RuntimeException;
 abstract class BaseFieldHandler implements FieldHandlerInterface
 {
     /**
-     * Default database field type
-     */
-    const DB_FIELD_TYPE = 'string';
-
-    /**
      * Flag for rendering value as is
      */
     const RENDER_PLAIN_VALUE = 'plain';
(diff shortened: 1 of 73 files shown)


src/FieldHandlers/BaseFieldHandler.php
```diff
@@ -389,6 +385,23 @@ public static function fieldToDb(CsvField $csvField)
...
 
+    /**
+     * Get database field type
+     *
+     * @param string $field Field name
+     * @return string
+     */
+    public static function getDbFieldType($field)
+    {
+        // Temporary dummy configuration
```
georgeconstantinou: @mamchenkov dummy configuration?

mamchenkov: @georgeconstantinou Yup.  This is for transitional stage only.  Both `fieldToDb()` and `getDbFieldType()` methods will stop being static soon.  Once they are not static, we can use the regular configuration instance.  For now though, we just get a dummy configuration instance with no relevance to any particular table, view, etc.

georgeconstantinou: I see

PR #523, opened 2018-02-09
Removal of individual field handlers (task #5061)

mamchenkov:
Removal all but one field handlers.

> One Handler to rule them all, One Handler to find them,
> One Handler to bring them all and in the darkness bind them


src/FieldHandlers/FieldHandlerFactory.php
```diff
@@ -177,15 +176,11 @@ public static function hasFieldHandler($fieldType)
      * @param  mixed  $table  name or instance of the Table
      * @return object         Table instance
      */
-    protected function _getTableInstance($table)
+    protected static function getTableInstance($table)
     {
         $tableName = '';
```
georgeconstantinou: @mamchenkov this variable is no longer used.

src/FieldHandlers/Provider/FieldToDb/CombinedFieldToDb.php
```diff
@@ -13,8 +13,8 @@
 
 use CsvMigrations\FieldHandlers\CsvField;
 use CsvMigrations\FieldHandlers\DbField;
+use CsvMigrations\FieldHandlers\FieldHandler;
```
georgeconstantinou: @alex-mamchenkov `DbField` and `FieldHandler` classes are not used here.

PR #559, opened 2018-05-31
Proposing Rich value list field handler (task #6106)

anvyst:
`RichList FieldHandler` allows embedding HTML values into `option` tags with the use of `select2` JS library.

It's an expansion of ListFieldHandler that has been used for JSON lists. It comes in handy for color-defined dropdown lists, and all sorts of HTML-based values for the select input field.

Example of `colors.json`:

```json
{
    "items": [
        {
          "value": "#F3C300",
          "label": "<div style='width:20px;height:20px;margin:0;float:left;background:#F3C300;'></div>Yellow",
          "inactive": null,
          "children": []
        },
        {
          "value": "#875692",
          "label": "<div style='width:20px;height:20px;float:left;background:#875692;'></div>Purple",
          "inactive": null,
          "children": []
        }
    ]
}
```

Example of `db/migration.json`:

```json

 "color": {
         "name": "color",
         "type": "rich_list(colors)",
         "required": null,
         "non-searchable": null,
         "unique": null
     }

```

### DIFF
--- a/src/FieldHandlers/Config/ConfigFactory.php
+++ b/src/FieldHandlers/Config/ConfigFactory.php
@@ -11,6 +11,7 @@
  */
 namespace CsvMigrations\FieldHandlers\Config;
 
+use Cake\Utility\Inflector;
 use RuntimeException;
 
 /**
@@ -33,7 +34,8 @@ class ConfigFactory
      */
     public static function getByType($type, $field, $table = null, array $options = [])
     {
-        $configClass = __NAMESPACE__ . '\\' . ucfirst($type) . 'Config';
+        $configClass = __NAMESPACE__ . '\\' . Inflector::camelize($type) . 'Config';
+
         if (!class_exists($configClass)) {
             throw new RuntimeException("Configuration type [$type] is not supported");
         }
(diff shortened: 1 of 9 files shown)


src/FieldHandlers/Config/ConfigFactory.php
```diff
@@ -33,7 +34,8 @@ class ConfigFactory
      */
     public static function getByType($type, $field, $table = null, array $options = [])
     {
-        $configClass = __NAMESPACE__ . '\\' . ucfirst($type) . 'Config';
+        $configClass = __NAMESPACE__ . '\\' . Inflector::camelize($type) . 'Config';
```
mamchenkov: @anvyst this creates the naming inconsistency with the other lists (Dblist, Sublist) ...

anvyst: @mamchenkov, it doesn't. Dblist and Sublist is not snake case, so it perfectly falls under Inflector rules, but using multiword type definition with snake case will break classical camel-case class notation - that's why `ucfirst()` was replace in favor or Inflector.

Example: `rich_list` will require `Rich_listConfig` class not `RichListConfig`.

PR #579, opened 2018-08-16
Record fetching query fixes (task #6717)

georgeconstantinou:
* Fetch by lookup fields only if the query string ID is not a UUID.
* Applying field alias for the primary key to avoid the following PDO exception:
```sql
Error: SQLSTATE[23000]: Integrity constraint violation: 1052 Column 'id' in where clause is ambiguous
```

### DIFF
--- a/src/Controller/AppController.php
+++ b/src/Controller/AppController.php
@@ -71,7 +71,7 @@ class AppController extends BaseController
     public function view($id = null)
     {
         $entity = $this->{$this->name}->find()
-            ->where([$this->{$this->name}->getPrimaryKey() => $id])
+            ->where([$this->{$this->name}->aliasField($this->{$this->name}->getPrimaryKey()) => $id])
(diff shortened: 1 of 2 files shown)


src/Controller/AppController.php
```diff
@@ -71,10 +72,10 @@ public function beforeFilter(Event $event)
     public function view($id = null)
     {
         $entity = $this->{$this->name}->find()
-            ->where([$this->{$this->name}->getPrimaryKey() => $id])
+            ->where([$this->{$this->name}->aliasField($this->{$this->name}->getPrimaryKey()) => $id])
             ->first();
 
-        if (empty($entity)) {
+        if (null === $entity && ! Validation::uuid($id)) {
```
mamchenkov: @georgeconstantinou I still prefer the `empty()` check to a much stricter null check here, but that's not critical :)

PR #587, opened 2018-08-31
Avatar for related records (task #6982)

softius:
Adds image_src if available before the link for related records (view only). Like avatar for users.

### DIFF
--- a/src/Template/Element/FieldHandlers/RelatedFieldHandler/view.ctp
+++ b/src/Template/Element/FieldHandlers/RelatedFieldHandler/view.ctp
@@ -10,20 +10,25 @@ foreach ($relatedProperties as $properties) {
     } else {
         // generate related record(s) html link
         $title = $properties['dispFieldVal'];
-        if (isset($properties['config']['table']['icon'])) {
+        $beforeLink = null;
+
+        // Special case for entities having an image_src like Users
+        if (isset($properties['entity']['image_src'])) {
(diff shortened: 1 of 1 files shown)


src/Template/Element/FieldHandlers/RelatedFieldHandler/view.ctp
```diff
@@ -10,19 +12,24 @@ foreach ($relatedProperties as $properties) {
     } else {
         // generate related record(s) html link
         $title = $properties['dispFieldVal'];
-        if (isset($properties['config']['table']['icon'])) {
+        $beforeLink = null;
+
+        // Special case for entities having an image_src like Users
+        if (Configure::read('Theme.prependAvatars', true) && isset($properties['entity']['image_src'])) {
```
mamchenkov: @softius I think `!empty()` is a better check than `isset()`.  The `image_src` can be set to an empty value, in which case the link shouldn't be rendered.

src/Template/Element/FieldHandlers/RelatedFieldHandler/view.ctp
```diff
@@ -10,19 +12,24 @@ foreach ($relatedProperties as $properties) {
...
+            $beforeLink = '<img alt="User Image" src="' . $properties['entity']['image_src'] . '" style="width: 20px; height: 20px;" class="img-circle"> ';
```
mamchenkov: @softius the `alt` attribute mentions the User image, but this might be slightly misleading as the entity might not necessarily be a User entity, but anything else with the `image_src`.  Maybe set it to something more generic like  'Image' or 'Thumbnail' or 'Preview'.

PR #593, opened 2018-09-07
[WIP] DataTable reload on record deletion (task #7074)

georgeconstantinou:
Fix for DataTable not reloading when a record is deleted. Additionally, the relevant DataTable will be reloaded whenever the related tab is clicked.

### DIFF
--- a/webroot/js/plugin.js
+++ b/webroot/js/plugin.js
@@ -22,7 +22,7 @@ $(document).ready(function () {
                 },
                 success: function (data) {
                     //traverse upwards on the tree to find table instance and reload it
-                    var table = $(hrefObj).closest('.table-datatable').DataTable();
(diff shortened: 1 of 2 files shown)


webroot/js/plugin.js
```diff
@@ -22,7 +22,7 @@ $(document).ready(function () {
                 },
                 success: function (data) {
                     //traverse upwards on the tree to find table instance and reload it
-                    var table = $(hrefObj).closest('.table-datatable').DataTable();
```
softius: table-datatable is being in some other places like ScheduledJobs, Dblists, Menus and Users

georgeconstantinou: @softius the `table-datatable` class is added by us, mostly for styling purposes, but is not always present (like in this case). `dataTable` class is always attached by the DataTables library, which makes it more consistent.

PR #610, opened 2018-09-24
Import shell test class (task #7273)

georgeconstantinou:


tests/TestCase/Shell/ImportShellTest.php
```diff
@@ -0,0 +1,106 @@
...
+     * @return void
+     */
+    public function testMain()
+    {
+        $table = TableRegistry::getTableLocator()->get('Articles');
+        $initialCount = $table->find()->count();
+
+        $this->ImportShell->main();
+
+        $this->assertNotEquals($initialCount, $table->find()->count());
```
mamchenkov: @georgeconstantinou maybe set a variable to the result of the `$table->find()->count()` instead of calling it separately for each assertion.

georgeconstantinou: @mamchenkov done: https://github.com/QoboLtd/cakephp-csv-migrations/pull/610/commits/95464a3a48f399fa8d6dbc3a4d978de477ec90fa

PR #611, opened 2018-09-24
Import PDO exceptions handling (task #7273)

georgeconstantinou:


src/Template/Import/progress.ctp
```diff
@@ -21,7 +21,7 @@ $statusLabels = [
     ImportsTable::STATUS_IN_PROGRESS => 'primary',
     ImportsTable::STATUS_COMPLETED => 'success',
     ImportsTable::STATUS_PENDING => 'warning',
-    ImportsTable::STATUS_FAIL => 'error'
+    ImportsTable::STATUS_FAIL => 'danger'
```
mamchenkov: @georgeconstantinou this will require a migration to update all existing records.  I think instead we should adjust the styling to use `danger` class, rather than change the status in the database.

src/Utility/Import.php
```diff
@@ -343,7 +343,7 @@ public static function setStatusLabels(array $data, $index)
         $statusLabels = [
             ImportResultsTable::STATUS_SUCCESS => 'success',
             ImportResultsTable::STATUS_PENDING => 'warning',
-            ImportResultsTable::STATUS_FAIL => 'error'
+            ImportResultsTable::STATUS_FAIL => 'danger'
```
mamchenkov: @georgeconstantinou same as above.

georgeconstantinou: @mamchenkov again, this is just for CSS class usage, it does not affect the database.

src/Template/Import/progress.ctp
georgeconstantinou: @mamchenkov this is just for CSS class usage, it does not affect the database.

PR #615, opened 2018-09-27
Aggregator Configuration class (task #6030)

georgeconstantinou:
This class is used as a storage object for all aggregator related configuration. Its constructor accepts the following arguments:
- Aggregated table instance
- Aggregated field name

Setter methods:
- `setDisplayField($fieldName)`
  - If **not** provided the aggregated field will be used as display field.
  - 
- `setJoinData($table, $entity)`
  - If used, the aggregated results will be limited to the provided entity's associated data.
  - The `$entity` argument is required to match the join table's entity class.

### Usage
Configuration for aggregating on "created" column of Articles table:
```php
use Cake\ORM\TableRegistry;
use CsvMigrations\Aggregator\Configuration;

$configuration = new Configuration(TableRegistry::get('Articles'), 'created');
```
Configuration for aggregating on `created` column of `Articles` table, limiting the dataset to the provided `author's` associated articles, and using `title` as display field:
```php
use Cake\ORM\TableRegistry;
use CsvMigrations\Aggregator\Configuration;

$table = TableRegistry::get('Authors');
$author = $table->find()->first();

$configuration = new Configuration(TableRegistry::get('Articles'), 'created');
$configuration->setDisplayField('title')
    ->setJoinData($table, $author)
```

For example of how to use the Configuration object with Aggregators see PR https://github.com/QoboLtd/cakephp-csv-migrations/pull/617.


softius:
@georgeconstantinou can you please also provide on how we can use the Configuration once it's created?

georgeconstantinou:
> @georgeconstantinou can you please also provide on how we can use the Configuration once it's created?

Sure, I will provide a link to the relevant PR.

mamchenkov:
@georgeconstantinou overall it looks fine to me, except the number of parameters that are needed to initialize the configuration.  Instead of making 4 or 5 parameters, what do you think about switching to the chain methods like `setTable()`, `setField()`, etc ... so the constructor will be simpler, and you can set the parameters in any order as you need ... and when you are actually using the configuration, you can throw an exception if a required parameter is missing .. just an idea :)

georgeconstantinou:
> @georgeconstantinou overall it looks fine to me, except the number of parameters that are needed to initialize the configuration. Instead of making 4 or 5 parameters, what do you think about switching to the chain methods like `setTable()`, `setField()`, etc ... so the constructor will be simpler, and you can set the parameters in any order as you need ... and when you are actually using the configuration, you can throw an exception if a required parameter is missing .. just an idea :)

@mamchenkov good point :+1:  the Configuration class has been adjusted now to provide setters for all optional properties (https://github.com/QoboLtd/cakephp-csv-migrations/pull/615/commits/5ac148ad23d7fd937a021326a333bf1ea85609d2)

PR #629, opened 2018-10-31
 FileUpload class cleanup (task #8246)

georgeconstantinou:
* Removed underscore prefix from all class properties and methods.
* Removed unused class constants.
* Moved static property to a constant.
* Better constant naming.
* Type-hinting to the interface.
* Added deprecated annotation to obsolete code.


src/Utility/FileUpload.php
```diff
@@ -16,73 +16,63 @@
...
-    protected $_fileStorageForeignKey;
+    protected $fileStorageAssociation;
 
     /**
      * Image file extensions
      *
      * @var array
      */
-    protected $_imgExtensions = ['jpg', 'png', 'jpeg', 'gif'];
+    protected $imgExtensions = ['jpg', 'png', 'jpeg', 'gif'];
```
mamchenkov: @georgeconstantinou I think we should have it somewhere in the `cakephp-utils` as this is used in many places throughout ... maybe not right now, but in the future ;)

PR #642, opened 2018-11-27
Add support for HTML fields (task #2668)

softius:
Introduces a new field handler for HTML. For this we are using the tinymce library.

In particular:
* TinyMCE editor is used when rendering the input
* HTML is included as-is when rendering the value
* HTML is stored in db as text
* Searching within HTML is supported

### DIFF
--- a/src/Template/Element/FieldHandlers/HtmlFieldHandler/input.ctp
+++ b/src/Template/Element/FieldHandlers/HtmlFieldHandler/input.ctp
@@ -23,39 +23,3 @@ $attributes += [
 
 echo $this->Form->control($name, $attributes);
 
-use Cake\Core\Configure;
-
-// load tinyMCE editor and elFinder file manager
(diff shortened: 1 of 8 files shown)


src/Template/Element/FieldHandlers/HtmlFieldHandler/input.ctp
```diff
@@ -0,0 +1,61 @@
...
+    'class' => 'tinymce' . ( (isset($extraClasses) && !empty($extraClasses)) ? ' ' . $extraClasses : null ),
+    'label' => $label,
+    'required' => (bool)$required,
+    'value' => $value,
+    'placeholder' => $placeholder,
+];
+
+echo $this->Form->control($name, $attributes);
+
+use Cake\Core\Configure;
```
mamchenkov: @softius can this be at the top please

PR #659, opened 2019-01-09
Implemented a currenciesFieldHandler (task #2653)

mitsios:
Created a new currencies field handler that adds the symbol in front and the description of the currency on the attribute title

### DIFF
--- a/config/currencies.php
+++ b/config/currencies.php
@@ -1,0 +1,431 @@
+<?php
+return [
+    'Currencies' => [
+        'list' => [
+            'ALL' => [
+                'description' => 'Albania Lek',
+                'symbol' => 'Lek'
+            ],
+            'AFN' => [
+                'description' => 'Afghanistan Afghani',
+                'symbol' => '؋'
+            ],
+            'ARS' => [
+                'description' => 'Argentina Peso',
+                'symbol' => '&#36;'
+            ],
+            'AWG' => [
+                'description' => 'Aruba Guilder',
+                'symbol' => 'ƒ'
+            ],
+            'AUD' => [
+                'description' => 'Australia Dollar',
+                'symbol' => '&#36;'
+            ],
+            'BSD' => [
+                'description' => 'Bahamas Dollar',
+                'symbol' => '&#36;'
+            ],
+            'BBD' => [
+                'description' => 'Barbados Dollar',
+                'symbol' => '&#36;'
+            ],
+            'BYN' => [
+                'description' => 'Belarus Ruble',
+                'symbol' => 'Br'
+            ],
+            'BZD' => [
+                'description' => 'Belize Dollar',
+                'symbol' => 'BZ&#36;'
+            ],
+            'BMD' => [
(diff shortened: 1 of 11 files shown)


src/FieldHandlers/Provider/RenderValue/CurrencyRenderer.php
```diff
@@ -0,0 +1,61 @@
...
+     * Provide rendered value
+     *
+     * @param mixed $data Data to use for provision
+     * @param array $options Options to use for provision
+     * @return mixed
+     */
+    public function provide($data = null, array $options = [])
+    {
+        $result = parent::provide($data, $options);
+        $hasError = sprintf(parent::VALUE_NOT_FOUND_HTML, $data);
```
softius: @mitsios `$hasError` is a bit confusing as a variable name ('has' denotes a boolean var)

src/FieldHandlers/Provider/CombinedFields/MoneyCombinedFields.php
```diff
@@ -27,7 +27,7 @@ class MoneyCombinedFields extends AbstractCombinedFields
             'config' => 'CsvMigrations\\FieldHandlers\\Config\\DecimalConfig'
         ],
         'currency' => [
-            'config' => 'CsvMigrations\\FieldHandlers\\Config\\ListConfig'
+            'config' => 'CsvMigrations\\FieldHandlers\\Config\\CurrencyConfig'
```
softius: @georgeconstantinou I am bit concerned here about existing implementations and how these will be affected... is there anyway to be backwards compatible with this change?

georgeconstantinou: @softius one way I can think is to add a configuration parameter and conditionally set the `$fields['currency']['config']` class in the class constructor.

mamchenkov: I think Money field handler should use the Currency, not the List.  We might need to reiterate over the Currency one a couple of times to find the most aesthetically pleasing way to render it, but I think that Currency is more correct than List.

config/currencies.php
```diff
@@ -0,0 +1,431 @@
...
+                'description' => 'Albania Lek',
+                'symbol' => 'Lek'
+            ],
+            'AFN' => [
+                'description' => 'Afghanistan Afghani',
+                'symbol' => '؋'
+            ],
+            'ARS' => [
+                'description' => 'Argentina Peso',
+                'symbol' => '&#36;'
```
mamchenkov: @mitsios do these display correctly in the dropdown lists? I think we had an issue before the HTML entities being escaped.

mitsios: yes, i noticed an issue while i was working on this ticket that that when the symbol was a dollar sign it was displaying some weird html in search page only, i tried to escape it or url encode it without success, so when i used the html code it work, and since a lot of currencies are using as symbol the dollar sign i replaced all of them with that html code and tested it. It worked

PR #662, opened 2019-01-15
Added a new commonmark render handler(task #2669)

mitsios:
Added a new handler for commonmark

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -31,7 +31,7 @@
         "eluceo/ical": "^0.11.0",
         "qobo/cakephp-translations": "^10.0",
         "qobo/cakephp-utils": "^11.0",
-        "league/commonmark": "^0.18.1",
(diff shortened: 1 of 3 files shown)


georgeconstantinou:
@mitsios can you also set `allow_unsafe_links` to `false`?
https://commonmark.thephpleague.com/security/#unsafe-links

composer.json
```diff
@@ -31,6 +31,7 @@
         "eluceo/ical": "^0.11.0",
         "qobo/cakephp-translations": "^10.0",
         "qobo/cakephp-utils": "^11.0",
+        "league/commonmark": "^0.18.1",
```
mamchenkov: @mitsios I think "^0.18" is a slightly better option as it will allow updates of bugfixes.

PR #670, opened 2019-02-13
Bug fix for select2 in embedded panel (task #9973)

eliamariutti:
`select2()` was not initialize correctly in the embedded FieldHandler

### DIFF
--- a/src/Template/Element/FieldHandlers/CountryFieldHandler/input.ctp
+++ b/src/Template/Element/FieldHandlers/CountryFieldHandler/input.ctp
@@ -10,25 +10,6 @@
  * @license       https://opensource.org/licenses/mit-license.php MIT License
  */
 
-$this->Html->script(
-    [
-        'AdminLTE./bower_components/select2/dist/js/select2.full.min',
-        'Qobo/Utils.select2.init',
-        'RolesCapabilities.switch-target',
-        'RolesCapabilities.permissions',
-    ],
-    ['block' => 'scriptBottom']
-);
-
-$this->Html->css(
-    [
-        'AdminLTE./bower_components/select2/dist/css/select2.min',
-        'Qobo/Utils.select2-bootstrap.min',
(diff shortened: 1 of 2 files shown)


src/Template/Element/FieldHandlers/CountryFieldHandler/input.ctp
```diff
@@ -10,25 +10,6 @@
...
-    [
-        'AdminLTE./bower_components/select2/dist/js/select2.full.min',
-        'Qobo/Utils.select2.init',
-        'RolesCapabilities.switch-target',
-        'RolesCapabilities.permissions',
-    ],
-    ['block' => 'scriptBottom']
-);
-
-$this->Html->css(
```
softius: @eliamariutti it seems that we have removed 7 assets files and introduced only one in common libs. Are the remaining assets not necessary  or being loaded somewhere else?

eliamariutti: Yes, all the `FiledHandler` has a `common_js_libs.ctp` with the js and css. 
I think is not necessary to load `RolesCapabilities` js for a country select.

src/Template/Element/common_js_libs.ctp
```diff
@@ -72,6 +72,7 @@ echo $this->Html->script(
         'CsvMigrations.plugin',
         'Qobo/Utils./plugins/tinymce/tinymce.min',
         'CsvMigrations.tinymce.init',
+        'Qobo/Utils.select2.init',
```
georgeconstantinou: @eliamariutti on line 71 we load select2.init file from CsvMigrations plugin, what are the differences between the two?

eliamariutti: @georgeconstantinou From what I have understood in the CsvMigrations plugin, manly we find the Select2 Logic while in the `Utils.select2.init` we just find the `$('.select2').select(...)` to makes everything work

PR #680, opened 2019-03-07
Self relation Module associations (task #10059)

eliamariutti:
Add support to M:M associations with the same table (self-referencing).

### DIFF
--- a/src/Model/AssociationsAwareTrait.php
+++ b/src/Model/AssociationsAwareTrait.php
@@ -20,6 +20,7 @@ use CsvMigrations\Utility\FileUpload;
 use Qobo\Utils\ModuleConfig\ConfigType;
 use Qobo\Utils\ModuleConfig\ModuleConfig;
 use Qobo\Utils\Utility;
+use RuntimeException;
 use Webmozart\Assert\Assert;
 
 /**
@@ -120,7 +121,7 @@ trait AssociationsAwareTrait
             return;
         }
 
-        $selfrelated = $this->isSelfRelated($fields);
+        $selfrelated = $this->getSelfRelated($fields);
         if (!empty($selfrelated)) {
             $this->setByTypeSelfRelationField($module, $selfrelated);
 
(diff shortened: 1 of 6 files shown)


src/Model/AssociationsAwareTrait.php
```diff
@@ -121,10 +121,35 @@ private function setByTypeRelation(string $module, array $fields) : void
         }
 
         foreach ($fields as $field) {
-            $this->setByTypeRelationField($module, $field, $moduleField);
+            $this->setByTypeRelationField($module, $field, $moduleField, $this->isSelfRelated($fields));
```
georgeconstantinou: @eliamariutti you can call `$this->isSelfRelated($fields)` before the _foreach_ loop, no need to be called for each field.

src/Model/AssociationsAwareTrait.php
```diff
@@ -120,36 +120,56 @@ private function setByTypeRelation(string $module, array $fields) : void
...
             $duplicate = $first;
         }
 
-        return false;
+        if (count($selfrelated) === 2) {
+            return $selfrelated;
+        }
+
+        if (count($selfrelated) > 2) {
+            throw new Exception('Related table with more than two fields');
```
georgeconstantinou: @eliamariutti can we change the message here to: _Many-to-many self-association with more than two fields is not supported._

Also is better to throw a `\RuntimeException` instead.

src/Model/AssociationsAwareTrait.php
```diff
@@ -120,11 +120,58 @@ private function setByTypeRelation(string $module, array $fields) : void
...
+    /**
+     * Check if the relation module is "self-related" : is true when
+     * two related fields point to the some module. In this case will
+     * be return an array with two CsvField object.
+     *
+     * @param  mixed[] $fields Module fields
+     * @return CsvMigrations\FieldHandlers\CsvField[]|null
+     * @throws Exception in case there are more than two fields pointing the some module.
+     */
+    private function isSelfRelated(array $fields)
```
softius: @eliamariutti @georgeconstantinou functions with `is` or `has` prefix is better to return boolean values. 

PS. dont forget the tests

georgeconstantinou: @softius true, @eliamariutti I think we need to change the method name to `getSelfRelated`.

src/Model/AssociationsAwareTrait.php
```diff
@@ -120,11 +121,57 @@ private function setByTypeRelation(string $module, array $fields) : void
...
+        }
+
+        if (count($selfrelated) === 2) {
+            return $selfrelated;
+        }
+
+        if (count($selfrelated) > 2) {
+            throw new RuntimeException('Many-to-many self-association with more than two fields is not supported.');
+        }
+    }
```
georgeconstantinou: @eliamariutti I think here you can return an empty array, for cases where the `selftrelated` count is less than 2. This way you can remove the nullable from the return type-hint and the void from the return doc-block.

PR #682, opened 2019-03-13
 Added a new manyToMany fieldHandler input (Task #10356)

mitsios:


src/Controller/AppController.php
```diff
@@ -12,10 +12,13 @@
 namespace CsvMigrations\Controller;
 
 use App\Controller\AppController as BaseController;
+use Assert\Assertion;
```
georgeconstantinou: @mitsios this class is not used.

src/FieldHandlers/Config/BelongsToManyConfig.php
```diff
@@ -0,0 +1,38 @@
...
+ * for HasMany field handlers.
+ */
+class BelongsToManyConfig extends FixedConfig
+{
+    /**
+     * @var array $providers List of provider names and classes
+     */
+    protected $providers = [
+        'combinedFields' => '\\CsvMigrations\\FieldHandlers\\Provider\\CombinedFields\\NullCombinedFields',
+        'dbFieldType' => '\\CsvMigrations\\FieldHandlers\\Provider\\DbFieldType\\UuidDbFieldType',
```
georgeconstantinou: @mitsios with this, if the field is added in `migration.json` it will create a new column in the database. You can create another class that will return an empty string, you can check _AggregatedDbFieldType_ as an example. In fact, we can rename the _AggregatedDbFieldType_ to _NullDbFieldType_ and share it between this config class and the aggregated config class.

PR #685, opened 2019-03-21
Improved positioning of calendar (task #10394)

stepanov:
Set property of the calendar "drops" to "up" or "down"  depends of the current offset of the element from the top - task #10394

### DIFF
--- a/webroot/js/datetimepicker.init.js
+++ b/webroot/js/datetimepicker.init.js
@@ -75,13 +75,15 @@
          * @return {object}
          */
         getOptions: function (input, args) {
+            let drops = $(input).offset().top > 360 ? 'up' : 'down;'
+
             var options = {
                 singleDatePicker: true,
                 showDropdowns: true,
                 timePicker: true,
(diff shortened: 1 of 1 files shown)


webroot/js/datetimepicker.init.js
```diff
@@ -75,13 +75,15 @@
          * @return {object}
          */
         getOptions: function (input, args) {
+            let drops = $(input).offset().top > 360 ? 'up' : 'down;'
```
georgeconstantinou: @stepanov can we inverse this so we have the down as default? For example check against `.bottom` instead of `.top`.

georgeconstantinou: Probably something like this: https://stackoverflow.com/questions/9872128/get-bottom-and-right-position-of-an-element?answertab=active#tab-top

anvyst: @georgeconstantinou it won't work for modal windows, looking for alternative solutions.

PR #686, opened 2019-03-26
Daterangepicker Magic Values (task #10535)

anvyst:
Daterangepicker disables `ranges` in case following flags are enabled:

* `timePicker` flag is `on`.
* `autoApply` is on and `ranges` object (not array) is not specified.

In our case, for DateTimeFieldHandler we use single Calendar layout,
and pre-populate calendar with magic values (today, yesterday, etc.),
thus we have to force showing custom ranges after daterangepicker
was constructed and magic values object was passed into constructor.

Reference: https://github.com/dangrossman/daterangepicker/blob/v2.1.27/daterangepicker.js#L372

### DIFF
--- a/webroot/js/datetimepicker.init.js
+++ b/webroot/js/datetimepicker.init.js
@@ -58,6 +58,11 @@
                 // date range picker (used for datetime fields)
                 $(this).daterangepicker(that.getOptions(this, options), that.getCallback(this, hiddenInput));
 
+
+                if ($(this).data('magic-value')) {
+                    $($(this).data('daterangepicker').container).find('.ranges ul').show();
+                }
+
                 $(this).on('apply.daterangepicker', function (ev, picker) {
                     $(this).val(picker.startDate.format(picker.locale.format));
                 });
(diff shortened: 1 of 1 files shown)


webroot/js/datetimepicker.init.js
```diff
@@ -58,6 +58,11 @@
                 // date range picker (used for datetime fields)
                 $(this).daterangepicker(that.getOptions(this, options), that.getCallback(this, hiddenInput));
 
+
+                if ($(this).data('magic-value')) {
+                    $($(this).data('daterangepicker').container).find('.ranges ul').show();
```
softius: @anvyst I am not sure about the `container` right after the `data('daterangepicker')` .... is this specific to `daterangepicker` ?

anvyst: @softius yep, it's internal property of the Daterangepicker:
https://github.com/dangrossman/daterangepicker/blob/v2.1.27/daterangepicker.js#L130

PR #706, opened 2019-07-04
Validation checks improvements (task #11809)

eliamariutti:
Creating an instance of the class where the json parsing are calculate only once, will improve the validation time drastically.

An example with a module with 270 fields in the view (before and after) :
`./bin/cake validate MyModule  125.27s user 0.10s system 99% cpu 2:05.39 total`
`./bin/cake validate MyModule  2.56s user 0.05s system 100% cpu 2.602 total`

### DIFF
--- a/src/Utility/Validate/Utility.php
+++ b/src/Utility/Validate/Utility.php
@@ -31,6 +31,50 @@ use Qobo\Utils\Utility\Convert;
  */
 class Utility
 {
+    protected static $configJsonArray;
+
+    protected static $migrationJsonArray;
+
+    /**
+     * We create a object that that set only once most of the json parsing
+     * @param string $module Module name
+     */
+    public function __construct(string $module = '')
+    {
+        if (empty($module)) {
+            return;
+        }
+
+        self::setConfigJsonArray($module);
+        self::setMigrationJsonArray($module);
+    }
+
+    /**
+     * Retrive and parse to array the config.json file
+     * @param  string $module Module name
+     * @param bool $validate Should the data be validated against the schema.
+     * @return void
+     */
+    public static function setConfigJsonArray($module, bool $validate = true)
+    {
+        $mc = new ModuleConfig(ConfigType::MODULE(), $module, null, ['cacheSkip' => true]);
+        $mc->setParser(new Parser($mc->createSchema(), ['validate' => $validate]));
+        self::$configJsonArray = Convert::objectToArray($mc->parse());
+    }
+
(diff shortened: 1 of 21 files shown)


tests/TestCase/FieldHandlers/TimeFieldHandlerTest.php
```diff
@@ -81,7 +81,7 @@ public function testRenderInputWithTimeObject() : void
 
     public function testRenderInputWithDateObject() : void
     {
-        $result = $this->fh->renderInput(new Date('13:30'));
+        $result = $this->fh->renderInput(new Date('2000-01-01 13:30'));
```
georgeconstantinou: @eliamariutti the value we use here for the test should represent what is being stored in the database. In the case of _time_ datatype columns, this will be something like `13:30:00`.

src/Utility/Validate/Utility.php
```diff
@@ -31,6 +31,52 @@
...
+     * Create an instance of the class where the json parsing are calculate only once
+     * @param string $module Module name
+     */
+    public function __construct(string $module = '')
+    {
+        if (empty($module)) {
+            return;
+        }
+
+        self::setConfigJsonArray($module);
```
softius: @eliamariutti I dont like that we mix non-static method with static setters. I believe you can either drop the constructor (can be replaced with another static method) or create a wrapper method.

src/Utility/Validate/Utility.php
```diff
@@ -123,10 +169,11 @@ public static function getRealModuleFields(string $module, bool $validate = true
     {
         $moduleFields = [];
 
-        $mc = new ModuleConfig(ConfigType::MIGRATION(), $module, null, ['cacheSkip' => true]);
-        $mc->setParser(new Parser($mc->createSchema(), ['validate' => $validate]));
-        $moduleFields = Convert::objectToArray($mc->parse());
-        $fields = Hash::extract($moduleFields, '{*}.name');
+        if (empty(self::$migrationJsonArray) && self::$module !== $module) {
```
softius: @eliamariutti since the getters will take care of the 'caching', i am suggesting the following:

* Introduce one more level in both `$migrationJsonArray` and `$configJsonArray` so that you hold the module name and multiple module fields ie. `$migrationJsonArray['ModuleName']` etc
* Remove `self::$module` and adjust accordingly
* Remove entirely `initStatic` as it would be no longer necessary (actions will be taken by the getters).

PR #709, opened 2019-08-06
Do not cast to string array fields (Task #12795)

acrias:
If the field is array it cannot be cast to string so we return and we do not proceed to render the field.

### DIFF
--- a/src/FieldHandlers/Provider/RenderValue/AbstractRenderer.php
+++ b/src/FieldHandlers/Provider/RenderValue/AbstractRenderer.php
@@ -32,7 +32,11 @@ abstract class AbstractRenderer extends AbstractProvider
     public function provide($data = null, array $options = [])
     {
         // TODO: need to handle this better, type-casting to string will produce errors
-        $result = (string)$data;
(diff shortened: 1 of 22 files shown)


src/FieldHandlers/Provider/RenderValue/AbstractRenderer.php
```diff
@@ -32,7 +32,11 @@ abstract class AbstractRenderer extends AbstractProvider
     public function provide($data = null, array $options = [])
     {
         // TODO: need to handle this better, type-casting to string will produce errors
```
eliamariutti: @acrias I think we can remove the comment

softius:
@acrias it would be really nice if you could expand the tests to cover this case as well.

src/FieldHandlers/Provider/RenderValue/AbstractRenderer.php
acrias: @eliamariutti  It has not finished yet, most probably we will come back again for more field types

PR #711, opened 2019-08-29
Timepicker bugfix (task #10976)

eliamariutti:
In case the project runs with the `CsvMigrations.tableValidation` enabled, the cakephp validation were failing due the wrong format of the hours in the time (*H* instead of *HH*). 
The `bootstrap-timepicker` JS library responsible to provide the time is no longer maintained and a small workaround is set to provide always two digit hours.

### DIFF
--- a/webroot/js/timepicker.init.js
+++ b/webroot/js/timepicker.init.js
@@ -41,11 +41,11 @@
             }
 
             // time picker
-            $('[data-provide="timepicker"]').timepicker(defaults).on('changeTime.timepicker show.timepicker', function(e) {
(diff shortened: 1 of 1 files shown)


webroot/js/timepicker.init.js
```diff
@@ -17,6 +17,7 @@
                 });
             });
         });
+        $("body").find('[data-provide="timepicker"]').data('timepicker').update()
```
georgeconstantinou: @eliamariutti do we need this?

eliamariutti: @georgeconstantinou Yes, the timepicker is responsible with the setTime function to crop the first digit of the hour. I could not find any other fast way to trigger the timepicker (so the changeTime) then this solution.

PR #717, opened 2019-09-23
Include css and js files when necessary depending on the FieldHandler (Task #10114)

acrias:
This PR removes many common libraries from loaded in common_js_libs.ctp and include them in the the FieldHandler files

### DIFF
--- a/src/Template/Element/FieldHandlers/FilesFieldHandler/value.ctp
+++ b/src/Template/Element/FieldHandlers/FilesFieldHandler/value.ctp
@@ -37,7 +37,7 @@ $limit = 3;
             <div class="thumbnail" title="<?= $entity->get('filename') ?>">
                 <?= $this->Html->image(Hash::get(
                     $entity->get('thumbnails'),
-                    in_array($entity->extension, FileUpload::IMAGE_EXTENSIONS) ? 'small' : 'large'
+                    in_array($entity->extension, FileUpload::IMAGE_EXTENSIONS) ? 'small' : 'huge'
                 )) ?>
                 <div class="caption">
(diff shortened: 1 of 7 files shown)


src/Template/Element/FieldHandlers/FilesFieldHandler/value.ctp
```diff
@@ -37,7 +37,7 @@ $limit = 3;
             <div class="thumbnail" title="<?= $entity->get('filename') ?>">
                 <?= $this->Html->image(Hash::get(
                     $entity->get('thumbnails'),
-                    in_array($entity->extension, FileUpload::IMAGE_EXTENSIONS) ? 'small' : 'large'
+                    in_array($entity->extension, FileUpload::IMAGE_EXTENSIONS) ? 'small' : 'huge'
```
georgeconstantinou: @acrias I think this was recently changed by @eliamariutti

acrias: @georgeconstantinou fixed

src/Template/Element/common_js_libs.ctp
```diff
@@ -15,15 +15,8 @@ use Cake\Core\Configure;
 echo $this->Html->css(
     [
         'AdminLTE./plugins/iCheck/all',
-        'Qobo/Utils./plugins/daterangepicker/css/daterangepicker',
```
softius: @acrias removing these assets / files will cause issues with template files including `common_js_libs` - for example https://github.com/QoboLtd/project-template-cakephp/search?q=common_js_libs&unscoped_q=common_js_libs

I would prefer to make the necessary adjustments so that this affects only this plugin while we keep `common_js_libs` as a deprecated file. I am thinking to introduce a new element for absolutely common file (like a base file).

PR #723, opened 2019-10-31
Transitions list (task #11973)

eliamariutti:
[Second PR] - Update JSON validation, pass the actual list status as option parameters.

List life-cycle refers to the possible transitions between list items, which are usually applied in fields like Status.

### DIFF
--- a/src/Utility/Validate/Utility.php
+++ b/src/Utility/Validate/Utility.php
@@ -105,7 +105,11 @@ class Utility
         $listItems = null;
         try {
             $mc = new ModuleConfig(ConfigType::LISTS(), $module, $list, ['cacheSkip' => true]);
+
+            $schema = $mc->createSchema(['lint' => true]);
+            $mc->setParser(new Parser($schema, ['lint' => true, 'validate' => true]));
             $config = $mc->parse();
+
             $listItems = property_exists($config, 'items') ? $config->items : null;
         } catch (InvalidArgumentException $e) {
             return false;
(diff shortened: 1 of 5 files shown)


src/Utility/Validate/Utility.php
```diff
@@ -114,6 +118,23 @@ public static function isValidList(string $list, string $module = '') : bool
         if (null === $listItems) {
             return false;
         }
+        $transitions = (array)Hash::extract($mc->parseToArray()['items'], '{n}.transitions');
```
georgeconstantinou: @eliamariutti no need to parse here is already done on line 111.

src/Utility/Validate/Utility.php
```diff
@@ -114,6 +118,23 @@ public static function isValidList(string $list, string $module = '') : bool
...
+
+        if (!empty($transitions)) {
+            $listKeys = array_flip((array)Hash::extract($mc->parseToArray()['items'], '{n}.value'));
```
georgeconstantinou: @eliamariutti same here, parsing already done on line 111.

src/Utility/Validate/Utility.php
```diff
@@ -115,6 +119,25 @@ public static function isValidList(string $list, string $module = '') : bool
             return false;
         }
 
+        $array = Convert::objectToArray($config);
+        $transitions = (array)Hash::extract($array['items'], '{n}.transitions');
+
+        if (!empty($transitions)) {
```
georgeconstantinou: @eliamariutti invert the condition here and do an early return, to avoid too much nesting.

PR #761, opened 2020-03-24
Proxy thumbnail generation via an event (task #15659)

asaliev:
Allows to manipulate what happens before and after the thumbnail generation

### DIFF
--- a/src/Event/ProcessThumbnailsListener.php
+++ b/src/Event/ProcessThumbnailsListener.php
@@ -1,0 +1,74 @@
+<?php
+
+/**
+ * Copyright (c) Qobo Ltd. (https://www.qobo.biz)
+ *
+ * Licensed under The MIT License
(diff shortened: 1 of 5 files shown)


src/Event/ProcessThumbnailsListener.php
```diff
@@ -0,0 +1,74 @@
...
+ * @copyright     Copyright (c) Qobo Ltd. (https://www.qobo.biz)
+ * @license       https://opensource.org/licenses/mit-license.php MIT License
+ */
+
+namespace CsvMigrations\Event;
+
+use Burzum\FileStorage\Model\Entity\FileStorage;
+use Cake\Event\Event;
+use Cake\Event\EventListenerInterface;
+use Cake\Log\LogTrait;
```
eliamariutti: @asaliev This is not used

PR #762, opened 2020-03-27
Fields helper render input updates (task #15480)

eliamariutti:
Implement a help tooltip for the input fields of the modules.
The helper text can be saved in the field.json file of the module.


webroot/js/datepicker.init.js
```diff
@@ -1,128 +1,3 @@
-(function ($) {
```
acrias: @eliamariutti Is this okay that we have removed the datepicker and entered a tooltip code?

eliamariutti: @acrias Nope, I overwrite the file accidentally. Nice catch!

src/View/AppView.php
```diff
@@ -32,6 +32,7 @@ public function initialize()
     {
         parent::initialize();
 
+        $this->loadHelper('MyHtml');
```
acrias: @eliamariutti the helper `MyHtml` couldn't be something else like `TooltipHelper`?

src/View/Helper/HelpHelper.php
```diff
@@ -0,0 +1,32 @@
...
+
+    /**
+     * Template for help tooltip
+     *
+     * @param string $message Help message
+     * @return string
+     */
+    public function help(string $message): string
+    {
+        return '<span data-toggle="tooltip" title="" class="badge bg-blue" data-placement="auto right" data-original-title="' . __($message) . '">?</span>';
```
acrias: @eliamariutti we could have this in a config and be able to adjust it if possible like the `class`, `title` and the `data-placement` for example

src/View/AppView.php
eliamariutti: @acrias I'm actually fixing here due is wrong how I load the class . The right code is 
`$this->loadHelper('Html', ['className' => 'MyHtml']);`
and the `MyHtml` it extends `Cake\View\Helper\HtmlHelper`. The class name is not important.

src/View/Helper/HelpHelper.php
eliamariutti: @acrias Sure we can, but this is kind of a standard tooltip. The html part is right to be inside the php class.

eliamariutti: I really like the idea to export the css class in a config file : but the `MyHtmlHelper` in the csv pluging is not actually used in the project-template because we load `App\View\AppView` instead of `CsvMigrations\View\AppView`. So I will do the change in the project-template PR [here](https://github.com/QoboLtd/project-template-cakephp/pull/959)

PR #770, opened 2020-05-11
Translate behavior in import (Task #16083)

eliamariutti:
Adding translations from a import file: during the import we check in the file headers if there are translated fields and update the entities using the Translate Behavior of CakePHP.


src/Shell/ImportShell.php
```diff
@@ -388,7 +390,41 @@ protected function _importResult(Import $import, array $headers, int $rowNumber,
...
+        foreach ($fields as $field => $value) {
+            if (!in_array($field, array_keys($data))) {
+                continue;
+            }
+
+            if (!method_exists($entity, 'translation')) {
+                throw new MissingBehaviorException("Translation behavior is not configured correctly: check TranslateTrait in the Entity class");
+            }
+
+            $entity->translation($value['lang'])->{$value['parent']} = $data[$field];
```
softius, line 423: @eliamariutti since `translation` returns an `Entity`, lets replace the above with `set()`

src/Shell/ImportShell.php
```diff
@@ -388,7 +390,41 @@ protected function _importResult(Import $import, array $headers, int $rowNumber,
...
+    /**
+     * Check for tranlsations, set the entity with Translate Behavior and update the data payload.
+     *
+     * @param Table $table Table class
+     * @param EntityInterface $entity New entity
+     * @param string[] $headers Upload file headers
+     * @param mixed[] $data Current data from file line
+     * @return mixed[]
+     */
+    protected function setLanguages(Table $table, EntityInterface $entity, array $headers, array $data): array
```
softius, line 407: @eliamariutti since objects are passed by ref, you shouldnt have any problem not returning the `$entity`. Ideally, you must return only the update data, and hence avoid using the `list` in caller methods.

PR #771, opened 2020-05-14
Import updates (task #16121)

eliamariutti:
Add the possibility to update records with using the import functionality.
The file needs to have an unique identifier to match the entity we want to update. If no record is found in the related model, the import will try to add it.

image

### DIFF
--- a/src/Template/Import/mapping.ctp
+++ b/src/Template/Import/mapping.ctp
@@ -107,6 +107,7 @@ $('select[name="options[options][update_identifier]"]').on("change", function(){
     <div class="row">
         <div class="col-md-10 col-lg-8">
             <div class="box box-primary">
+                <?php if(!empty($identifier)): ?>
                 <?= $this->Form->create($import, ["id" => "mapping"]) ?>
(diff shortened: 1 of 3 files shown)


src/Template/Import/mapping.ctp
```diff
@@ -86,8 +107,38 @@ echo $this->Html->scriptBlock(
...
+                <?php if(!empty($identifier)): ?>
+                <?= $this->Form->create($import, ["id" => "mapping"]) ?>
                 <div class="box-body">
-                <?= $this->Form->create($import) ?>
+                <div class="row">
+                    <div class="col-md-3"></div>
+                    <div class="col-md-8">
+                        <div class="form-group input checkbox">
+                            <input type="checkbox" name="options[options][update]" value="1" class="square" id="is_update">
+                            <label class="control-label">Update matching records.</label>
```
softius, line 118: @eliamariutti this one must be wrapped with `__d`.

acrias:
@eliamariutti do we have any pot file?

PR #782, opened 2020-06-02
Update currency config (task #16520)

eliamariutti:


config/currencies.php
```diff
@@ -6,6 +6,10 @@
                 'description' => 'Albania Lek',
                 'symbol' => 'Lek',
             ],
+            'AED' => [
+                'description' => 'UAE Dirham',
+                'symbol' => 'د.إ',
```
acrias, line 11: We could have the `&#1583; .&#1573;`

config/currencies.php
```diff
@@ -386,6 +390,10 @@
                 'description' => 'Trinidad and Tobago Dollar',
                 'symbol' => 'TT&#36;',
             ],
+            'TRY' => [
+                'description' => 'Turkish Lira',
+                'symbol' => '₺',
```
acrias, line 395: And here the `&#8378; &#x20BA;`

config/currencies.php
softius: @acrias we keep non encoded values here on purpose, to avoid encoding when rendering the value.

PR #784, opened 2020-06-04
Filter uploads by mime (task #16581)

eliamariutti:
Allow upload by mime type list (backend)

### DIFF
--- a/config/file_upload.php
+++ b/config/file_upload.php
@@ -1,0 +1,19 @@
+<?php
+
+return [
+    'FileUpload' => [
+        'allowedMime' => [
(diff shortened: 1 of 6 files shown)


config/file_upload.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+return [
+    'FileUpload' => [
```
acrias, line 4: Do we want to allow all those mime types? Are we using anything else except images and pdfs?

eliamariutti: @acrias The idea is to allow everything from a plugin level : in the App, we will overwrite the list as needed.

PR #785, opened 2020-06-15
Disable dynamic panel in batch edit (task #16689)

eliamariutti:
The dynamic panel in the batch edit has all the field enabled when appear the form.
This will reset all the information hold in this panel for each entity selected in the batch.

### DIFF
--- a/webroot/js/panels.js
+++ b/webroot/js/panels.js
@@ -78,7 +78,9 @@
     };
 
     Panel.prototype.resetPanels = function () {
-        $(this.form).find('[data-provide="dynamic-panel"].hidden').find(':input').attr('disabled', false);
+        if (window.location.href.indexOf("/batch/edit") === -1) {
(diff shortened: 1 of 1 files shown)


webroot/js/panels.js
```diff
@@ -78,7 +78,9 @@
     };
 
     Panel.prototype.resetPanels = function () {
-        $(this.form).find('[data-provide="dynamic-panel"].hidden').find(':input').attr('disabled', false);
+        if (window.location.href.indexOf("/batch/edit") === -1) {
```
softius, line 81: @eliamariutti any idea on how we can make the URL here more dynamic?

eliamariutti: @softius Not really. My point of view is that that URL is not dynamic in any possible way and should stay a const as it is now. Any idea is welcome.

PR #788, opened 2020-06-18
Allow sorting files on specific fields (Task #10489)

acrias:
This PR enables the file sorting on specific file fields that have orderBy on field `order` in the fields.json


config/csv_migrations.php
```diff
@@ -74,11 +74,12 @@
                 'showUploadedThumbs' => false,
                 'reversePreviewOrder' => false,
                 'uploadAsync' => true,
-                'dropZoneEnabled' => true,
+                'dropZoneEnabled' => false,
```
softius, line 77: @acrias I am bit concerned that the defaults have been changed here. Can you please advise?

src/Template/Element/FieldHandlers/FilesFieldHandler/input.ctp
```diff
@@ -10,14 +10,23 @@
...
 $class = str_replace('.', '_', $name . '_ids');
 
+$config = ModuleRegistry::getModule($table)->getFields();
+
 $options = [
     'multiple' => true,
     'data-upload-url' => sprintf("/api/%s/upload", $table),
 ];
+
+if (isset($config[$field]['orderBy']) && 'order' == $config[$field]['orderBy'] && isset($config[$field]['orderDir'])) {
```
softius, line 26: @acrias lets move `order` into config so that we remove this coupling (at least) from project-template-cakephp.

config/csv_migrations.php
acrias: @softius this was already false before and I have changed to true and then back to false. Actually it doesn't make any change for these values as we already changing them in the javascript file.

src/Template/Element/FieldHandlers/FilesFieldHandler/input.ctp
acrias: @softius how this will be used in config

softius: @acrias in `config` we will need to define the field to be used for sorting, similar to how behavior work.

PR #791, opened 2020-06-25
Limit number of files to be uploads (Task #16681)

acrias:
This PR allow you to limit the number of files allowed to be uploaded for each field

### DIFF
--- a/webroot/js/fileinput-load.js
+++ b/webroot/js/fileinput-load.js
@@ -166,9 +166,12 @@ $(document).ready(function () {
             showDrag = true;
         }
 
+        var maxFileCountAllowed = $(inputField).data('file-limit');
+
         var existing = {
             showUpload: false,
             showCaption: true,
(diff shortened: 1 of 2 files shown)


webroot/js/fileinput-load.js
```diff
@@ -315,10 +318,13 @@ $(document).ready(function () {
         if (1 == $(inputField).attr('data-file-order')) {
             showDrag = true;
         }
+
+        var maxFileCountAllowed = $(inputField).data('file-limit');
         // Keep existing images on adding new images,
         // overwrtting default options in case of existing files
         var existing = {
             showUpload: false,
+            maxFileCount: maxFileCountAllowed,
```
softius, line 327: @acrias we get unlimited file uploads, when the limit is not set right?

acrias: @softius if you do not specify any limit the files that can be uploaded are unlimited

PR #797, opened 2020-07-08
Allow virtual fields in index and view files (task #16980)

asaliev:


src/Utility/Validate/Check/ViewsCheck.php
```diff
@@ -217,9 +217,14 @@ protected function getModuleConfig(string $module, string $view, array $options
         $mc = new ModuleConfig(ConfigType::VIEW(), $module, $view, ['cacheSkip' => true]);
 
         $schema = $mc->createSchema(['lint' => true]);
-        $schema->setCallback(function (array $schema) use ($module) {
+        $schema->setCallback(function (array $schema) use ($module, $view) {
             //Add fields from migration file
             $schema = $this->addFieldsToSchema($schema, $module);
+
+            // Add virtual fields to `index` and `view`
+            if (in_array($view, ['index', 'view'])) {
```
acrias, line 225: @asaliev by doing this meaning that is not required to use the virtualfield property in config.json? Because for the index is currently working

asaliev: @acrias in order for things to work this is necessary + changes from utils.

PR #799, opened 2020-07-09
[WIP]Allow virtual fields in index and view files (task #16980)

acrias:
Back-port changes from master to allow virtual fields on index and view files


acrias:
We do not want this anymore under v38

PR #803, opened 2020-08-18
Allow importing sublist fields via importShell (Task #17299)

acrias:
This PR allow importing values for sublist fields in the format of <parent name>.<child name>. The values should be converted to parent_name.child_name


src/Shell/ImportShell.php
```diff
@@ -485,6 +485,9 @@ protected function _processData(Table $table, array $csvFields, array $data): ar
                     case 'list':
                         $data[$field] = $this->_findListValue($table, $csvFields[$field]->getLimit(), $value);
                         break;
+                    case 'sublist':
```
softius, line 488: @acrias we need to implement a reverse lookup for this one. Like the one that happens in `_findListValue`

PR #809, opened 2020-09-30
Adjust type attribute to consider MS Office Applications as objects

acrias:
This PR fixes an issue with fileinput plugin that MS Office Applications were considered as images


src/Template/Element/FieldHandlers/FilesFieldHandler/input.ctp
```diff
@@ -37,13 +37,37 @@ if (isset($config[$field]['limit'])) {
 if ($value && $entities && $entities->count()) {
     $options['data-document-id'] = $value;
     $files = [];
+
     foreach ($entities as $entity) {
+        switch ($entity->mime_type) {
+            case 'application/pdf':
```
softius, line 43: @acrias can we move this into config? Something very simple where key is the `mime_type` and value the preview type

softius (changes requested):

softius (approved):

PR #822, opened 2020-10-26
New simple units fields (Task #13661)

eliamariutti:
New Currency, Distance and Area fields with the option to select currency/metric through settings.


webroot/js/inputmask.init.js
```diff
@@ -0,0 +1,3 @@
+$(document).ready(function () {
+    $(':input').inputmask()
```
acrias, line 2: @eliamariutti I think this must be called only for those inputs that has `data-inputmask`

acrias (changes requested):

webroot/js/inputmask.init.js
eliamariutti: @acrias it is. Check `src/Template/Element/FieldHandlers/SimpleUnitFieldHandler/input.ctp`

acrias (approved):

softius (approved):

PR #834, opened 2021-01-29
DBlist render input does not return correct results (Task #18860)

acrias:
This PR fixes an issue with renderInput of DBlistRenderer that the values returned were re-numbered


nicosp (commented):
What does this change?

eliamariutti (approved):

nicosp (approved):

PR #838, opened 2021-03-27
Pass selected field name to lookupActionListener

acrias:
This PR passes the form field name to lookupActionListener so that we know which field is it

### DIFF
--- a/webroot/js/select2.init.js
+++ b/webroot/js/select2.init.js
@@ -82,6 +82,7 @@ var csv_migrations_select2 = csv_migrations_select2 || {};
             ajax: {
                 url: function() {
 
+                    var currentFieldName = $(input).attr('field-name');
                     var connectedField = $(input).data('connected-field') ?? '';
                     var connectedTargetField = $(input).data('connected-target-field') ?? '';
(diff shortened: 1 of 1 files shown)


webroot/js/select2.init.js
```diff
@@ -82,14 +82,21 @@ var csv_migrations_select2 = csv_migrations_select2 || {};
...
+                    var currentFieldName = $(input).attr('field-name');
                     var connectedField = $(input).data('connected-field') ?? '';
                     var connectedTargetField = $(input).data('connected-target-field') ?? '';
                     var connectedFieldValue = '';
                     if (connectedField != '' && connectedTargetField != '') {
                         connectedFieldValue = '?' + connectedTargetField + '=' + $('#' + connectedField).val();
                     }
 
-                    return $(input).data('url') + connectedFieldValue;
+                    if (connectedFieldValue != '' && currentFieldName != '') {
```
softius, line 93: @acrias param method could be handy here https://api.jquery.com/jquery.param/ to avoid messing with `&` and `?`

softius (approved):

PR #840, opened 2021-04-19
Enable auto increment field update (Task #19581)

acrias:
This PR enable us to update a missing auto increment field value during a record update

### DIFF
--- a/src/Event/Model/AutoIncrementEventListener.php
+++ b/src/Event/Model/AutoIncrementEventListener.php
@@ -64,18 +64,12 @@ class AutoIncrementEventListener implements EventListenerInterface
             return;
         }
 
-        // skip modifying auto-increment field(s) on existing records.
-        if (! $entity->isNew()) {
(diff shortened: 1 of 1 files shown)


src/Event/Model/AutoIncrementEventListener.php
```diff
@@ -64,16 +64,13 @@ public function autoIncrementFieldValue(Event $event, EntityInterface $entity, A
             return;
         }
 
-        // skip modifying auto-increment field(s) on existing records.
-        if (! $entity->isNew()) {
-            foreach (array_keys($fields) as $field) {
-                Assert::isInstanceOf($entity, Entity::class);
```
acrias, line 70: @nicosp @softius am I missing something here having the Assert in the loop instead outside the loop?

softius: it should be out, as the loop has nothing to do with the Entity.

softius (approved):

PR #846, opened 2021-05-13
[WIP]Update date and datetime validators

acrias:
Use the new validators (cake 3.8+)

### DIFF
--- a/src/FieldHandlers/Provider/ValidationRules/DateValidationRules.php
+++ b/src/FieldHandlers/Provider/ValidationRules/DateValidationRules.php
@@ -13,6 +13,9 @@
 
 namespace CsvMigrations\FieldHandlers\Provider\ValidationRules;
 
+use Cake\Validation\Validator;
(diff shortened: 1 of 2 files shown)


acrias:
@nicosp when trying to send this value to a reminder field via api I get an error `A validation error occurred`. This error appears also before changing the code. This throws a validation error `2021-05-13T10:10:10+03:00` but this is not throwing an error `2021-05-13T10:10:10`. @softius what the correct value that we should allow here

softius:
@acrias valid values are defined by the chosen validator and the corresponding CakePHP ORM Type. Any value should be fine as long as it validates and gets marshaled to a Date/DateTime object.

acrias:
> @acrias valid values are defined by the chosen validator and the corresponding CakePHP ORM Type. Any value should be fine as long as it validates and gets marshaled to a Date/DateTime object.

nicosp (approved):

